PR #64, opened 2023-04-05
optimize dockerfile

nathanosdev:
This PR makes a few changes to the Dockerfile.

First change fixes an issue with building the Dockerfile on M1 Macs.

I also added a `.dockerignore` file, which will prevent `node_modules` and `target` from being copied into the image, which I've realized were contributing to the build times. With the `.dockerignore` file the "load build context" and "COPY . ." steps are quicker but the "make build" step takes longer. That tells me that the `target` directory was being copied in and Rust was using it as a cache. I think it's better to not use those files from the host, but I'm not particularly opinionated about it.

Last change is downloading a prebuilt binary for `ic-wasm` instead of using `cargo install` to install `ic-cdk-optimizer`. `cargo install` will always build from source so it takes quite a while to build. `ic-cdk-optimizer` was deprecated in favor of `ic-wasm` but the end result is actually 120kb _bigger_ than `ic-cdk-optimizer`. I'm gonna check with the team to see if this is expected. But downloading the prebuilt binary saves quite a bit of time. It should also work to download the prebuilt binary for `ic-cdk-optimizer` if its preferred to stick with that for now, let me know and I can make that change.

Before:
image

After:
image

After note: the CI pipeline was about 5 mins faster.


taggrx:
Thanks so much @nathanosdev! 🙏 This is amazing, thanks so much for taking care of this! Please let me know once you checked with the team, why the produced binary has increased in size. However, we're far from it being a problem, so if ic-cdk-optimizer is deprecated, I think we should go with ic-wasm!

taggrx (approved):

nathanosdev:
No problems at all! I checked with the SDK team and they're aware of the increase in file size, it's because `ic-cdk-optimizer` is stripping out custom sections that `ic-wasm` doesn't. It's on their roadmap to improve this later.

PR #65, opened 2023-04-06
Release

taggrx:
#Release Changelog (🟢 low risk, high importance)

We should *not* let the fear paralyze us! Remeber, it wasn't a bug that took Taggr down. We're stronger than ever and will get more resilient with the next couple of upgrades. This is a low risk update and contains only quick fixes and some things I want to do as a first step after resurrection, to finally have a piece of mind.

- Restores stalwarts by setting the active weeks to 0 (we'll restore it later).
- Transfers 6k tokens to @christian and 600 to @dsarlis for their help getting Taggr back. Tokens are not minted! They are transferred from my _own wallet_ as one time exception. I hope to incentivize them to stay involved into #Taggr which is beneficial to all of you.
- Fixes the chores timer which is currently spamming the logs.
- Fixes the logout for II users.
- Fixes a minor UI issue.
- Restricts a user controller to a 1-1 relation (security related, found during the downtime).


src/backend/lib.rs
```diff
@@ -72,6 +74,43 @@ fn post_upgrade() {
...
+    // Thanks for implementing the replica fix! :praying_hands:
+    let dsarlis = "v7jwe-qzysn-aoxfx-ptfm2-54no7-csest-eay3s-utdaw-btja6-bycj4-bae";
+    token::transfer(
+        time(),
+        state_mut(),
+        x,
+        TransferArgs {
+            from_subaccount: None,
+            to: account(Principal::from_text(christian).unwrap()),
+            amount: 600_000,
```
chmllr, line 93: Wow, thank you, are you sure? This is almost too generous 😅

chmllr (approved):

src/backend/lib.rs
taggrx: My tokens, my decision 😎

PR #66, opened 2023-04-07
Low-level recovery

taggrx:
# #Release Notes: low-level recovery (🔴 HIGH RISK)

While we were offline and I had a lot of time to reflect I realized that the current recovery method we have (a proposal for assigning a canister controller) has severe disadvantages because it relies upon too many mechanisms to works:
  - post creation cannot fail
  - proposal creation cannot fail
  - voting cannot fail
  - proposal execution cannot fail.

Another huge disadvantage is that it still allows only a centralized recovery, which _only_ allows the assigned controller to upload a new binary.

In this MR I completely remove this proposal type and introduce a new redundant upgrade capabilities to #Taggr, which does not reduce the decentralization by a single bit but does not rely on any of the mechanisms above. Instead, it allows any stalwart to upload an emergency binary under `#/recovery` and allows users with token balances to submit a hash of the emergency binary and hence confirming the emergency upgrade. Once we reach the normal approval rate, the canister will upgrade itself. This method provides a simplified UX and has no automated voting power reduction.

## Other changes

- make memory sync fail-safe, however since this allocator is not used yet, these changes are not relevant yet.
- trim post contents to remove unused white space
- minor improvements and styling bugfixes.


taggrx:
Guys, the expectation is of course that you review only to the degree it makes sense for you. No worries, I am not offloading the responsibility, I care the whole responsibility myself. I only want to give a fair chance to everyone to spot potential problems in an upgrade.

chmllr (approved):
LGTM as far as I am able to judge.

recovery.md
radu-daniel, line 1: What to review here? :)

mechaquan:
How does it work if the main canister is down?

We need redundancy and recovery which doesn't jeopardise decentralization, right?!

I really think a separate wallet dapp canister which holds the $Taggr tokens and is rarely updated but used to login to main Taggr canister and the two canisters control each other could really be an interesting option.  

Or the wallet could just be used to login into a recovery canister as well as main prod canister.  The principals for each would be linked to the wallet, right?

You could still have a wallet on main dapp for normies but they would be advised to get a wallet setup.

mechaquan:
> LGTM as far as I am able to judge.

What is LGTM?!  Let's Go To Market?!

radu-daniel (dismissed):

taggrx:
> 

LGTM is a abbreviation in software engineering, especially on reviews and means "looks good to me".

taggrx:
> How does it work if the main canister is down?

It doesn't. This is a recovery for a situation where a bug prevents further upgrades becasue one of the 4 mechanisms listed above fails deterministically (post creation, proposal creation, standard voting, execution, etc).

> We need redundancy and recovery which doesn't jeopardise decentralization, right?!

Ideally yes. But I don't know how to solve this problem yet. 

> I really think a separate wallet dapp canister which holds the $Taggr tokens and is rarely updated but used to login to main Taggr canister and the two canisters control each other could really be an interesting option.

I think we got stuck on the question how to keep user logins in sync? But anyway, this is not related in any way to these code changes. This is only a replacement for the proposal adding a controller to the canister. This is not a recovery for all cases.

> Or the wallet could just be used to login into a recovery canister as well as main prod canister. The principals for each would be linked to the wallet, right?

Yes, we could have a separate canister where we login first, then we would switch to the actual Taggr canister and somehow link the login. Then the Taggr canister would regularly push all user updates to the login canister. In this case we wouldn't need any second canister, as the main one can be upgraded from the login canister. This will be a massive project. E.g. I don't know exactly right now how to do this session linking. I'm sure it's possible, but I'd need to investigate. But this is a different topic and should be discussed on Taggr.

> You could still have a wallet on main dapp for normies but they would be advised to get a wallet setup.

What if they don't? Is the whole upgrade power reduced to only those who managed to set it up? I don't think this is secure, so the linking shouldn't be optional imo.

nathanosdev:
>I think we got stuck on the question how to keep user logins in sync?

I believe II's alternative origins can do this. [The docs](https://internetcomputer.org/docs/current/developer-docs/integrations/internet-identity/alternative-origins) list _"Configuring multiple apps in your organization to use the same principals"_ as one of the uses cases, although I haven't actually seen it used like this yet so I have some uncertainty there.

I haven't spent too much time thinking about this though so there could be many things that I'm not considering, but I think splitting out a ledger canister in addition to a governance canister would be good. Governance would be executed on the governance canister but the UI for creating proposals and casting votes could be shared between the governance canister and the primary canister. So you can use the primary canister for convenience but in case of an emergency, the governance canister has the same UI that everyone can also log into. Since the ledger is split from both canisters there shouldn't be any issues with replicating tokens either. Of course, that assumes that alternative origins will work the way I think.

For the current PR, creating an independent path for emergency recoveries that is not affected by the more commonly used code paths makes sense. There's a downside to this though in my opinion because it (hopefully) won't be used very often so it _could_ break and it's possible nobody would notice until someone actually tried to use it. I think we need to be very careful in reviewing the code of `confirm_emergency_release` and any functions that it calls in the future to make sure that this doesn't break unexpectedly. There's not that much code though to be fair, and I think anyone would be cautiously reviewing the `upgrade_main_canister` and `active_voting_power` functions anyway. It would be nice to have some e2e automation tests in the future covering this scenario.

nathanosdev (approved):

taggrx:
Thanks so much @nathanosdev for looking into the MR! 

> I believe II's alternative origins can do this. [The docs](https://internetcomputer.org/docs/current/developer-docs/integrations/internet-identity/alternative-origins) list _"Configuring multiple apps in your organization to use the same principals"_ as one of the uses cases, although I haven't actually seen it used like this yet so I have some uncertainty there.

This is already used on Taggr for all domains. The problem with this method is [this](https://internetcomputer.org/docs/current/developer-docs/integrations/internet-identity/alternative-origins#listing-origins). The original canister has to be online and serve a list of derivation domains, authenticating them this way. But when the main canister is down, it obviously doesn't work. One could of course create a _yet_ another canister which would simply sit around and never do anything except providing the list of derivations for II users, but honestly I'd rather vote for removing the II login as impractical login method instead.

> For the current PR, creating an independent path for emergency recoveries that is not affected by the more commonly used code paths makes sense. There's a downside to this though in my opinion because it (hopefully) won't be used very often so it _could_ break and it's possible nobody would notice until someone actually tried to use it. I think we need to be very careful in reviewing the code of `confirm_emergency_release` and any functions that it calls in the future to make sure that this doesn't break unexpectedly. There's not that much code though to be fair, and I think anyone would be cautiously reviewing the `upgrade_main_canister` and `active_voting_power` functions anyway. It would be nice to have some e2e automation tests in the future covering this scenario.

You're absolutely correct in every point here. I do plan to
- almost never change emergency code (this is the point of extracting the critical and moving parts from its control flow)
- regularly test the emergency upgrade as it is really a normal upgrade with the only disadvantage of poor UX.
- invite other people for reviews whenever I touch this code or any code I consider as high risk.

Thanks a gain for your feedback, it's very important and valuable!

PR #67, opened 2023-04-09
Bugfixes

taggrx:
# #Release Changelog: bugfixes and improvements (🟢 LOW RISK)

- Fixes a bug in principal change and adds a fixture which transfers funds of 3 affected users to subaccounts controlled by the new principals.
- Realm theme fixes.
- More consisten token balances computation.
- Fixes meta information in raw assets (for crawlers).


src/frontend/src/index.jsx
```diff
@@ -26,7 +26,11 @@ import {Tokens, Transaction} from "./tokens";
 import {Whitepaper} from "./whitepaper";
 import {Recovery} from "./recovery";
 
-const { protocol, host, pathname } = location;
+const { hash, search } = location;
+
+if (!hash && search) {
+    location.href = `#${search.slice(1)}`;
```
nathanosdev, line 32: Are the crawlers requesting the path as a query instead of a hash fragment?

src/backend/assets.rs
```diff
@@ -105,7 +105,7 @@ fn add_asset(paths: &[&str], headers: Headers, bytes: Vec<u8>) {
 }
 
 pub fn asset(path: &str, certified: bool) -> Option<(Headers, ByteBuf)> {
-    let (headers, bytes) = assets().get(path)?;
+    let (headers, bytes) = assets().get(path).or_else(|| assets().get("/"))?;
```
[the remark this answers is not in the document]
taggrx: Yep, you're missing context and I didn't provide it.

Before out letdown we had a nginx serving paths like `share.taggr.link/post/123` by simply redirecting them to the `raw` urls. Then, the http router would recognize the path, fail to find asset for `/post/123` and simply serve an index.html but with some `og:` metadata for crawlers. So this path was never executed for certified assets and we never added certification due to the check in line 110. Nevertheless I think your remarkis fair and there is no way why the fallback should be explicit and on the upper level like it is right now for certified assets. I added a small refactoring commit.

src/frontend/src/index.jsx
taggrx: So as I explained in the previous comment we had the nginx which we don't have anymore. What I need to achieve though - assuming that the DFINITY's service worker will route requests to raw - is to recognize which path was requested. Since we're using hashes for routing on the frontend, the url like `taggr.links/#/post/123` will only be visible as `/` in the canister when the http query is transmitted. So the trick I plan to use is to use URLs `taggr.links/?/post/123` as share links, because the search part (after `?`) is passed to the http queries and I can simply again serve an index.html filled with the right meta information.

This code is speculative and is not rolled out as a feature yet. I just see no way how to test this stuff with the local `dfx`. Do you know it maybe?

taggrx:
@nathanosdev @chmllr  Since this MR contains urgent fixes and is low risk anyway, I'm creating a release from it, but obviously only vote for it, if you couldn't spot any new bugs.

nathanosdev:
Thanks for the extra context @taggrx, the changes make a lot more sense now.
Regarding testing the SEO type stuff locally there's no ready made solution for this yet. We use a Docker setup to try and emulate a simple boundary node when testing the service worker locally: https://github.com/dfinity/ic/tree/master/typescript/service-worker
Something like this could also work for canisters in general but it would involve a lot of copy/pasting and manually keeping the service worker up to date. I think we could try and distribute a Docker image to simplify this, but I could also try throw together a slight variation for Taggr to use now.

PR #71, opened 2023-04-12
NNS proposals fetch + voting + weighted polls

chmllr:
# Functionality

The new logic added in this PR: once per hour fetch NNS proposals and if there are new governance or replica election proposals, create a new post with a poll. (@taggrx remember to add Taggr as a controller of @IcBot — that's what discussed, right?) Then check hourly for polls that expire. All NNS proposals are created with a 3 days deadline, that means Taggr will vote in 3 days on each proposal using the votes on the poll weighted by the square root of karma.

## Notes

- The code in `canisters.rs` is almost entirely copied from the backend canister of my telegram bot. The voting code is new but this is fairly straightforward.
- I used the same neuron id as we used for the motion proposal, @taggrx please replace with whatever neuron you would like to use in the initial testing phase and do not forget to add Taggr canister as a hot key to this neuron!
- I added "hourly chores" by simply copying the logic from the existing `chores` routine. The hourly routine just checks if we have polls that are expired and computes the weights by user karma and tokens. The second step of this routine is testing if any of the expired polls being an NNS proposal. If yes, vote for the proposal.
- The function `conclude_poll` and the corresponding test were the hardest parts to figure out how to do. The test works, so I think it's at least not very wrong.
- I did not _touch any UI_ as I need to refresh my JS skills first. So maybe in the next PR we can start displaying the weighted poll results on UI?
- This code is low to medium risk I'd say as all changes are additive, are triggered once per hour and do not affect any functionality. In the worst case the neuron will not vote yet.


mechaquan:
So is this just a standard poll to start?

chmllr:
> So is this just a standard poll to start?

> So is this just a standard poll to start?

Not sure I understand the question. It’s the only poll we have. I only added weights to votes given to different options so that we know which option leads by karma or tokens. This additional info can be displayed later on UI. But it’s not part of this PR yet. 

I’m not sure which other polls we need?

src/backend/env/mod.rs
```diff
@@ -984,7 +1005,103 @@ impl State {
         user_ids
     }
 
+    fn conclude_polls(&mut self, now: u64) {
+        for post_id in self.pending_polls.clone() {
```
nathanosdev, line 1009: Instead of cloning here, the interface for `conclude_poll` could be updated to expect a reference for `post_id`

src/backend/env/mod.rs
```diff
@@ -984,7 +1005,103 @@ impl State {
...
+    async fn hourly_chores(&mut self, now: u64) {
+        if now - self.last_hourly_chores < HOUR {
+            return;
+        }
+        self.last_hourly_chores += HOUR;
+
+        self.conclude_polls(now);
+
+        // Vote on proposals if pending ones exist
+        for (proposal_id, post_id) in self.pending_nns_proposals.clone() {
```
nathanosdev, line 1030: Similarly here, we can use references instead of cloning

src/backend/env/post.rs
```diff
@@ -410,6 +416,51 @@ pub async fn add(
...
+        poll.weighted_by_karma = poll
+            .votes
+            .clone()
+            .into_iter()
+            .map(|(k, ids)| {
+                (
+                    k,
+                    ids.iter()
+                        .filter_map(|id| users.get(id))
+                        .map(|user| (user.karma() as f32).sqrt() as Karma)
```
nathanosdev, line 439: Could this be `filter_map(|id| users.get(id).map(|user| (user.karma() as f32).sqrt() as Karma)))` to save iterating over the array twice?

nathanosdev (approved):
LGTM! 👍

mechaquan:
> Not sure I understand the question. It’s the only poll we have. I only added weights to votes given to different options so that we know which option leads by karma or tokens. This additional info can be displayed later on UI. But it’s not part of this PR yet. 

Yeah that was my question...I didn't know if weighting had been implemented yet and if it was just going to be a standard 1 vote : 1 account.

So if I understand correctly weighting is implemented but not yet seen on UI?!

src/backend/env/post.rs
weedpatch2, line 439: Could we end up in a situation where a user has negative Karma when this is executed? Is that handled elsewhere?

weedpatch2 (approved):
Looks excellent!

src/backend/env/mod.rs
chmllr: The cloning is necessary to "cut ties" with `pending_polls`, because it's borrowed mutably inside the loop.

src/backend/env/post.rs
chmllr: There are no explicit iterations in Rust. This flow is iterator based and all steps are collapsed by the compiler into a single one before we continue with processing of the next element. In this concrete case, you simply pull the second `.map` inside the closure of `.filter_map`, but the step execution will still happen if and only if the user could be resolved. So we don't save anything, but compromise on readability. Also, from my own experience, whatever inlining or code reshuffling an engineer thinks would be more optimal are anyway done by the compiler optimisations and probably in a much more efficient way... 😅

src/backend/env/mod.rs
nathanosdev: Ah I see... makes sense now, thanks!

src/backend/env/post.rs
nathanosdev: Interesting... I didn't know about this, thanks for explaining!

src/backend/env/post.rs
```diff
@@ -410,6 +416,53 @@ pub async fn add(
...
+                    ids.iter()
+                        .filter_map(|id| users.get(id))
+                        .filter_map(|user| balances.get(&account(user.principal)))
+                        .sum(),
+                )
+            })
+            .collect();
+
+        poll.result_by_karma = result_by_karma;
+        poll.result_by_tokens = result_by_tokens;
```
mechaquan: So are you running both results by karma and by tokens to determine the best or most representative result?

My question/concern is someone who has earned a lot of karma but sold tokens (I know this isn't possible yet but should be resilient code, right?) or someone who has never earned karma but bought lots of tokens.  Or even someone who earned karma/tokens but them was reported and lost significant karma.

So I wonder if to minimize giving either of those scenarios a disproportionate vote weighting what if it's the average of karma + tokens squared?

chmllr: But IIUC @taggrx was against using any TAGGR tokens VP on NNS proposals?

As a first step I just computed both weights — one "ethical' from karma and one "economical" from tokens. We can't use both on the same poll of course. For the NNS proposals we agreed to use the ethical weighting. In the next release — I guess — we need to introduce a possibility on a poll selecting which weighting to apply to a particular poll so that we can see which option has won in the UI. I imagine for all DAO decisions we might want to go with the economical weighting to be in sync with the VP on proposals. Otherwise, a poll computed from both karma _and_ tokens might deviate from the sentiment of the DAO on the actual upgrade proposal which is computed from tokens only...

I guess we can introduce a third option you are suggesting which is a mix of both for something else. But shouldn't we discuss this on Taggr first?

src/backend/env/canisters.rs
```diff
@@ -219,6 +223,103 @@ pub async fn top_up(canister_id: Principal, min_cycle_balance: u64) -> Result<bo
...
+        pub proposer: Option<NeuronId>,
+        pub proposal: Option<ProposalStruct>,
+        pub topic: i32,
+    }
+
+    let args = ListProposalInfo {
+        include_reward_status: Default::default(),
+        before_proposal: Default::default(),
+        limit: 15,
+        exclude_topic: vec![2 /* ExchangeRate*/],
```
taggrx, line 269: Didn't you propose to vote on all proposal? But I agree that voting on the exchange rate proposals is pointless. Should the neuron follow DFINITY on them?

src/backend/env/canisters.rs
```diff
@@ -219,6 +223,103 @@ pub async fn top_up(canister_id: Principal, min_cycle_balance: u64) -> Result<bo
...
+        include_status: Default::default(),
+    };
+    let (response,): (ListProposalInfoResponse,) =
+        call_canister(MAINNET_GOVERNANCE_CANISTER_ID, "list_proposals", (args,))
+            .await
+            .expect("couldn't call governance canister");
```
taggrx, line 275: Let's have a `Result` here instead like in `vote_on_nns_proposal`?

src/backend/env/canisters.rs
```diff
@@ -219,6 +223,103 @@ pub async fn top_up(canister_id: Principal, min_cycle_balance: u64) -> Result<bo
...
+
+    response
+        .proposal_info
+        .into_iter()
+        .filter_map(|info| info.proposal.clone().map(|p| (info, p)))
+        .map(|(i, p)| NNSProposal {
+            id: i.id.unwrap().id,
```
taggrx, line 282: How about an `except` instead?

src/backend/env/mod.rs
```diff
@@ -984,7 +1005,103 @@ impl State {
...
+                Ok(post_id) => {
+                    self.pending_nns_proposals.insert(proposal.id, post_id);
+                }
+            }
+            self.last_nns_proposal = proposal.id;
+        }
+    }
+
     pub async fn chores(&mut self, now: u64) {
+        self.hourly_chores(now).await;
```
taggrx, line 1103: I'd move out the rest of `chores` into daily and weekly chores.

taggrx (approved):

taggrx:
Christian thanks so much! 🙏 The neuron id is fine — I have no other neuron anyway and we can indeed test with that one.

src/backend/env/canisters.rs
chmllr: They will be removed now any moment btw.

src/backend/env/canisters.rs
chmllr: Actually I can just `unwrap_or_default`!

src/backend/env/mod.rs
chmllr: Yep, I can clean up.

taggrx (approved):

PR #72, opened 2023-04-13
add intial e2e test boilerplate

nathanosdev:
This PR adds the minimum boilerplate for running e2e tests with Playwright.
I've also added a really simple sanity check spec that asserts the home page loads after deploying with DFX.
And I had to make some tiny Rust changes to keep Clippy happy, apparently patch updates include breaking changes now 🙄


README.md
```diff
@@ -36,3 +36,13 @@ Use the printed code commit and the binary to submit a new release proposal.
    - As a stalwart, execute `api.call("heap_to_stable")` in the the browser console.
    - Locally run: `./backup.sh /path/to/backup`.
 9. Restore the backup to the local canister: `./backup.sh /path/to/backup restore`.
+
+## Running e2e Tests
```
nathanosdev, line 40: I'll merge this with the changes from my other PR for docs depending on which one is merged first: https://github.com/TaggrNetwork/taggr/pull/69

playwright.config.ts
```diff
@@ -0,0 +1,25 @@
...
+  fullyParallel: true,
+  forbidOnly: !!process.env.CI,
+  retries: process.env.CI ? 2 : 0,
+  workers: process.env.CI ? 1 : undefined,
+  reporter: "html",
+  use: {
+    baseURL: "http://127.0.0.1:8080",
+    trace: "on-first-retry",
+  },
+  projects: [
```
nathanosdev, line 14: Only Chrome is tested for now, but we can easily add other browsers and screen sizes if we think it will be useful.

.dockerignore
```diff
@@ -17,3 +17,8 @@ target/
 node_modules/
 dist/
 public
+
+# e2e tests
+/test-results/
+/playwright-report/
+/playwright/.cache/
```
chmllr, line 24: Why global paths actually?

Makefile
```diff
@@ -24,6 +24,9 @@ fe:
 	rm -rf ./dist ./public
 	npm run build
 
+e2e_test:
```
chmllr, line 27: Should we make it part of `make build` like we execute `cargo test` there?

chmllr (dismissed):
LGTM!

.dockerignore
nathanosdev: That was what the Playwright init script added and I don't think that we'll have nested folders with the same name, but I'm happy to make that change just in case.

Makefile
nathanosdev: `e2e_test` has a dependency on `make build` so we wouldn't be able to add it there.

The order of instructions needs to be:
- `make build`
- `make start`
- `make dev_deploy`
- `make stop` (although this doesn't actually exist yet)

We could definitely add an `All in one` command that does all of this. Maybe it could be called `build_and_test` and then `cargo test` could also be moved to that instruction as well to keep `build` as just build.

What do you think?

.dockerignore
chmllr: Oh, it's the Docker file, I think this makes sense then.

Makefile
nathanosdev: It's already part of the Github CI :) https://github.com/TaggrNetwork/taggr/actions/runs/4726016583/jobs/8385123042?pr=72

I'll put together an all-in-one command later this evening

e2e/tests/sanity_check.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { test, expect } from "@playwright/test";
+import { HomePage } from "../pages";
+
+test("home page loads", async ({ page }) => {
+  const homePage = new HomePage(page);
+
+  await homePage.goto();
+
+  await expect(homePage.welcomeAboardHeader).toHaveText("Welcome aboard");
```
taggrx, line 9: Nice!!!

playwright.config.ts
taggrx: I think it's perfectly fine as a start!

taggrx (dismissed):
Thanks so much Nathan!

taggrx:
Wow, that's amazing! What was the issue btw.?

nathanosdev:
> What was the issue btw.?

I was the issue 😆 I was trying to make DFX working during the image creation phase of Docker. As soon as I moved it to the entrypoint, it worked.

and yes plz review and test 🙏

Makefile
```diff
@@ -16,16 +16,25 @@ build:
...
+	make start || true # don't fail if DFX is already running
+	make dev_deploy
+	npx playwright test
+	dfx stop
+
 release:
 	docker build -t taggr .
-	docker run --rm --entrypoint cat taggr /target/wasm32-unknown-unknown/release/taggr.wasm.gz > release.wasm.gz
-	shasum -a 256 release.wasm.gz
+	docker run -v $(shell pwd)/release-artifacts:/target/wasm32-unknown-unknown/release taggr
```
nathanosdev, line 38: I removed the `--rm` while debugging something, I'll add that back this evening.

release.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
```
taggrx, line 1: Could you please explain what exactly is happening by running this file as an entry point? Is `make build` still happening within the docker environment? Because if not, is the build even still a reproducible one?

nathanosdev: Yes, `make build` still runs within the Docker environment, it's just running within a different "phase" of the Docker image's lifecycle.

Previously, it was run during the "image building" phase. Now it happens when the image is "run", i.e. when it becomes a "container". Previously the only thing that happened during the "container" phase was extracting the file that was built during the "image building" phase, this was the `cat` command that you had set up as the entry point.

The reason to move `make build` to the "container" phase now is so we can make use of Docker volumes to extract the file instead of the `cat` entry point. If we want to run the e2e tests in the container, then we can't use `cat` as the entry point anymore.

If you want to double check the reproducibility yourself, the hash that I'm getting consistently locally when I run it multiple times across 2 machines (and a 3rd virtual machine, just for good measure) is `e25ee0532deb0cc71c0bbc21cf25138dbe950c08c9ca4e4802ecabb067b4c3e2`.

taggrx: Great thanks so much!

Btw, feel free to open PRs in this repo. You are member of the org and don't need to clone the repo.

PR #78, opened 2023-04-17
Realm clean up

taggrx:
As discussed on Taggr this PR implements a possibility for a realm clean up (I'd like to clean the NNS-GOV realm before we start advertizing with the introduction of the named neuron). This is a very low-risk code.


radu-daniel (dismissed):

mechaquan:
What do you by clean up exactly?

taggrx:
> What do you by clean up exactly?

The realm controller will see an additional button on every post, which will remove the post from the realm and penalize with `10` points the poster. This is intended to be user for posts which are too benign for reporting but do not fit into realm because users didn't read the rules or just don't know what they're doing.

WHITEPAPER.md
```diff
@@ -46,7 +46,11 @@ If after the bootcamp period the user still has less than `$trusted_user_min_kar
 ## Realms
 
 Realms represent sub-communities grouped by a certain topic.
-Every realm can have its own term and conditions, so that any violation of them can lead to flagging of user's post to stalwarts.
+Every realm can have its own term and conditions, so that any violation of them can lead to
+
+- flagging of user's post to stalwarts.
```
chmllr, line 51:
```suggestion
- flagging of user's post to stalwarts,
```

src/backend/env/mod.rs
```diff
@@ -212,6 +209,30 @@ pub enum Destination {
...
+        let post = self.posts.get(&post_id).ok_or("no post found")?;
+        let realm = post.realm.as_ref().ok_or("no realm id found")?;
+        if !post
+            .realm
+            .as_ref()
+            .and_then(|realm_id| self.realms.get(realm_id))
+            .map(|realm| realm.controllers.contains(&controller))
+            .unwrap_or_default()
+        {
+            return Err("only realm controllercan clean up".into());
```
chmllr, line 223:
```suggestion
            return Err("only realm controller can clean up".into());
```

src/backend/env/mod.rs
```diff
@@ -212,6 +209,30 @@ pub enum Destination {
...
+        }
+        let user = self.users.get_mut(&post.user).ok_or("no user found")?;
+        let msg = format!("post {} was moved out of realm {}", post_id, realm);
+        user.change_karma(-(CONFIG.realm_cleanup_penalty as Karma), &msg);
+        let user_id = user.id;
+        let penalty = CONFIG.realm_cleanup_penalty.min(user.cycles());
+        self.charge(user_id, penalty, msg)
+            .expect("couldn't charge user");
```
chmllr, line 231: But should we panic? We wouldn't be able to move out posts of some body who created a lot of junk and then run out of cycles.

chmllr (approved):

src/backend/env/mod.rs
taggrx: Thanks!

src/backend/env/mod.rs
taggrx: Well this panic is impossible, since in line 229 I already account for this case, right?

chmllr: True! 😅

taggrx:
> Will the post be deleted or reposted to main Taggr?

It simply gets moved out of the realm to Taggr.

PR #79, opened 2023-04-17
Bugfixes

taggrx:
- Fixes hash tags recognition on posts with CUTs and another small issue,
- Fixes polls,
- Removes the V1 seed phrase login, but introduces a temporary fixture which can be triggered by @X changing principals for all those users who didn't read the migration notice and will complain that they can't login,
- Increases the stalwart minimal activity to 2 weeks.


src/frontend/src/content.jsx
```diff
@@ -70,8 +72,9 @@ export const Content = ({post, value = "", blobs = [], collapse, preview, primeM
 
 const isALink = val => val.match(/^https?:\/\/.+$/) || val.match(/^www\..+$/);
 
-const markdownizer = (value, urls, setUrls, blobs, primeMode, className = null) => 
-    React.useMemo(() => !value ? null : <ReactMarkdown children={value} remarkPlugins={[remarkGfm]} className={className}
+const markdownizer = (value, urls, setUrls, blobs, primeMode, className = null) => !value
+    ? null
+    : <ReactMarkdown children={value} remarkPlugins={[remarkGfm]} className={className}
```
nathanosdev, line 77: Everything looks good, but I wasn't able to identify what issue is being solved here. Could you add some context?

taggrx: Thanks for paying attention!

The short answer is: try to create a comment on Taggr with three empty lines.

It's a memoization I added very long time ago to avoid Youtube flickering when writing comments. It never worked and now it also breaks the comments...

nathanosdev: That actually happened to me earlier 😆 but it happened while creating a post in a Realm. Thanks for the extra context.

nathanosdev (approved):

PR #81, opened 2023-04-21
Last endorsements

taggrx:
To increase transparency and make the karma farming harder, this new module on user's profile will show last endorsements of a user broken down by percentages.

Screenshot 2023-04-21 at 10 57 36

Another small changes:

- increase stalwart activity to 3 weeks
- 3 small UI bugfixes
- cleanups


mechaquan:
I think you are that mechaquan guy and karma farming!  But seriously this is cool and I need to stop upvoting you LOL!

src/backend/env/post.rs
```diff
@@ -211,11 +211,13 @@ impl Post {
...
         }
         // negative reactions balance
+        let karma = reaction_karma();
         if self
             .reactions
             .iter()
-            .filter_map(|(id, _)| CONFIG.reactions.iter().find(|(rid, _)| id == rid))
-            .map(|(_, cost)| *cost)
+            .map(|(r_id, users)| {
+                karma.get(r_id).copied().unwrap_or_default() * users.len() as Karma
```
chmllr, line 219: It looks like a changed logic. The multiplication was missing before?

src/backend/lib.rs
```diff
@@ -743,6 +733,23 @@ fn user() {
...
+                .rev()
+                .take(CONFIG.feed_page_size * 3)
+                .filter_map(|post_id| state.posts.get(post_id))
+                .flat_map(|post| post.reactions.iter())
+                .flat_map(|(r_id, users)| {
+                    let cost = karma.get(r_id).copied().unwrap_or_default();
+                    users.iter().map(move |user_id| (*user_id, cost))
+                })
+                .fold(BTreeMap::default(), |mut acc, (user_id, karma)| {
+                    acc.entry(user_id).and_modify(|e| *e += karma).or_insert(1);
```
chmllr, line 750: I think this is a bug
```suggestion
                    acc.entry(user_id).and_modify(|e| *e += karma).or_insert(karma);
```

chmllr (approved):
Looking forward to my top endorsements! 😅

src/backend/env/config.rs
```diff
@@ -184,7 +186,7 @@ pub const CONFIG: &Config = &Config {
 
     stalwart_percentage: 2,
     // TODO: set back to 6 in 6 weeks
-    min_stalwart_activity_weeks: 2,
+    min_stalwart_activity_weeks: 3,
```
mechaquan, line 189: Is the plan to walk this up to 4?  Should we change the % back to 3?  There are currently no spots...

nathanosdev:
This is a very cool idea! But I find the phrase "last endorsements" a bit confusing. What does "last" mean?
Does this apply to someone's total karma or just new karma?

What do you think of "karma sources"? Or "new karma source" if this applies only to new karma.

Will dig deeper into the code tonight, but it looks like Christian has it covered 👍

src/backend/env/post.rs
taggrx: Yes, it was a minor bug before which I discovered. When we compute the karma balance we of course need to multiple with the number of users... 😅

src/backend/lib.rs
taggrx: Oh, thanks so much! I'm so glad we have the reviews now. In my initial implementation I only counted the endorsers. Then I decided to count the actual karma and forgot to update this line. Thank you!!!

src/backend/env/config.rs
taggrx: That's a great idea! Will increase to 3%.

PS: the activity will go up to back to 6 as the TODO suggests. But we can discuss this parameter as we like.

taggrx:
> This is a very cool idea! But I find the phrase "last endorsements" a bit confusing. What does "last" mean? Does this apply to someone's total karma or just new karma?

Haha believe or not I struggled with the title a lot... These endorsements are taken from the last 90 posts of a user. So first I had "recent endorsements" but then they appear of course on a profile of a user inactive for a long time too, which makes the word "recent" kind of wrong.

"Karma sources" sounds too technical to me.

Maybe "Karma from latest posts"?

nathanosdev:
>Maybe "Karma from latest posts"?
That sounds good to me 👍 or maybe even "Karma from last 90 posts", I imagine a lot of people will ask what "latest" means.

mechaquan:
> >Maybe "Karma from latest posts"?
> 
> 
> 
> That sounds good to me 👍 or maybe even "Karma from last 90 posts", I imagine a lot of people will ask what "latest" means.

Love the directness 😂 skip the marketing and jump right to exactly what it is!

src/backend/env/user.rs
```diff
@@ -54,6 +54,9 @@ pub struct User {
     pub active_weeks: u32,
     pub principal: Principal,
     pub report: Option<Report>,
+
+    #[serde(default)]
+    pub endorsements: BTreeMap<UserId, Karma>,
```
nathanosdev, line 59: Maybe in light of the UI change, this could be updated to `karma_from_last_posts` so it's easier to relate to what's in the UI. Similar for related variable names in other files.

I think promoting a consistent language across the UI and the code base will help future developers understand what's going on.

src/frontend/src/profile.jsx
```diff
@@ -42,6 +43,9 @@ export const Profile = ({handle}) => {
     const { feed_page_size } = backendCache.config;
     const user = api._user;
     const showReport = profile.report && !profile.report.closed && user && user.stalwart;
+    const endorsements = Object.keys(profile.endorsements).map(id => [id, profile.endorsements[id]]);
```
nathanosdev, line 46: [Object.entries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) does the same thing as `Object.keys(value).map(key => [key, value[key]])`.

src/frontend/src/profile.jsx
```diff
@@ -54,8 +58,37 @@ export const Profile = ({handle}) => {
...
+                    {<UserLink id={userId} />}
+                    <code>{percentage(karma, endorsementsTotal)}</code>
+                </div>)}
+            </div>
+            {!allEndorsememnts && <button className="top_spaced" onClick={() => setAllEndorsements(true)}>SHOW ALL</button>}
+        </div>
+        <hr />
+        {profile.accounting.length > 0 && <>
+            <div className="spaced">
+                <h1>Karma and Cycles Changes</h1>
```
nathanosdev, line 75: To be "semantically correct" html, the `h1` should refer to the title of the page.
I think all of the existing headings should be sibling headings, so all `h2`.

Having "Karma and Cycles Changes" as `h1` and then "Latest Posts" as `h2` will imply to a screenreader that "Latest Posts" is somehow a "child" of "Karma and Cycles Changes", but this isn't the case.

If you want to visually make "Karma and Cycles Changes" bigger than the others, then I think it would be better to do this with CSS classes. Then the semantic structure of the HTML can be separate from the styling.

src/frontend/src/profile.jsx
```diff
@@ -54,8 +58,37 @@ export const Profile = ({handle}) => {
             </div>} />
         {showReport && <ReportBanner id={profile.id} reportArg={profile.report} domain="misbehaviour" />}
         <UserInfo profile={profile} />
-        {trusted(profile) && !stalwart(profile) && !isBot(profile) && <>
+        <div className="spaced">
+            <h2>Karma from last {backendCache.config.feed_page_size * 3} posts</h2>
+            <div className={bigScreen() ? "four_column_grid" : "two_column_grid"}>
```
nathanosdev, line 63: I'm not suggesting to make this change now, but just something to think about for the future.
Changing the number of columns in a CSS grid based on screen size is pretty trivial with media queries. It would reduce the markup complexity and also allow the design to adapt based on the screen size, which can be useful when you're resizing your window or changing orientation from portrait to landscape on a tablet device. Did you have any particular reason to drive this originally with JS?

nathanosdev (dismissed):

src/backend/env/user.rs
taggrx: Agreed, done.

src/frontend/src/profile.jsx
taggrx: I always forget about that one. Thanks so much! Will fix.

src/frontend/src/profile.jsx
taggrx: You are right. It started with JS because I needed it once. I refactored it.

src/frontend/src/profile.jsx
taggrx: I didn't even know that! Thanks so much, changed!

bhawanb:
That's great 👍

PR #83, opened 2023-04-21
Accept posts without realms while being in a realm

taggrx:
This change is needed for the IcBot so that it can post to NNS-GOV realm and outside of it (for whale alerts e.g.). This can be achieved by entering the NNS-GOV realm by the bot. Then the whale alert posts wouldn't specify any realm, which would make them go outside of the currently enetered realm.


src/backend/env/mod.rs
```diff
@@ -2594,7 +2594,7 @@ pub(crate) mod tests {
             p1,
             0,
             None,
-            None,
+            Some("SYNAPSE".into()),
```
chmllr, line 2597: Synapse name in tests? 😅

chmllr (approved):

src/backend/env/mod.rs
taggrx: This was implemented when they decided to use Taggr :)

PR #84, opened 2023-04-22
add signup with internet identity e2e test

nathanosdev:
This PR adds an e2e test that signs up for a new account on Taggr using Internet Identity.
It includes creating a new Internet Identity anchor and transferring ICP to the Taggr ICP account using a locally installed ledger and then minting cycles and setting a new username.

In a future PR I'll convert this test to a [Playwright fixture](https://playwright.dev/docs/test-fixtures) that can be used by other tests so the login/signup logic won't need to be repeated for other tests that need it. I think it would be useful to also create a fixture for signup/login with seed phrase. Then we can easily run each test with both authentication methods.

The test is a bit flaky, but I've opened a PR on agent-js to address this: https://github.com/dfinity/agent-js/pull/714.
I think we should wait until agent-js is updated and released before merging this PR.

What was happening here, is the agent-js was calling `onSuccess` before it was finished writing auth data to storage and then Taggr was reloading the page before that writing was complete. The auth initialization logic that was trigger on page load in Taggr then didn't have access to the auth data that was never fully written to storage.

This is definitely a fault of agent-js, but it did also bring my attention to the reloading pattern of Taggr. Which is unexpected for a single-page application. I think we should try to move towards a more "reactive" architecture on the frontend and I have some ideas for this, but I think I should finish the e2e tests first. I'll come with a proposal to the DAO for how to address this.

P.S. I officially love Playwright now. There's a lot of async stuff happening during this flow, changing routes, page reloads, component re-renders, state changes, opening new tabs, closing them and switching back to the main tab and I only had to use one "wait for" operation and I never had to re-query a selector. In other Selenium based frameworks the code for a flow like this would be absolutely littered with "wait for" operations and re-queries.


src/frontend/src/header.jsx
```diff
@@ -32,7 +32,7 @@ export const Header = ({subtle, route}) => {
                         toggler={() => setShowLogins(!showLogins)} currState={() => showLogins} onLabel="CLOSE" offLabel="CONNECT" />}
                 </>}
                 {api._principalId && 
-                    <BurgerButton onClick={() => { toggleButtonBar(!showButtonBar); toggleRealms(false) }} pressed={showButtonBar} />}
+                    <span data-testid="burger-button"><BurgerButton onClick={() => { toggleButtonBar(!showButtonBar); toggleRealms(false) }} pressed={showButtonBar} /></span>}
```
nathanosdev, line 35: I'm not a fan of doing this and I avoid it at all costs, but in these cases I had no other option since the DOM structure didn't lend any anchors that I could easily target from the e2e tests.

taggrx: But why don't you just pull the `data-testid` attribute inside the `BurgerButton`, why do you wrap instead?

Same question for other cases.

release.sh
```diff
@@ -1,7 +1,23 @@
...
+EOF
+dfx start --clean --background
+dfx nns install
+dfx stop
+
 make build
 make start
 make dev_deploy
-npx playwright test
+PWTEST_SKIP_TEST_OUTPUT=1 npx playwright test
```
taggrx, line 22: What's happening here?

taggrx (commented):
I wonder in general how this work. Is the II canister installed with the rest of NNS canisters?

taggrx:
Haha you even found a bug in agent-js, you rock! 😂

Another question: why didn't you start with the much easier seed phrase login?

release.sh
nathanosdev: When Playwright tests fail they start running a web server to display a report. The web server keeps the Docker container alive and prevents it from stopping. This env variable tells Playwright not to run that report, even if the tests fail: https://github.com/microsoft/playwright/issues/9702

nathanosdev:
>why didn't you start with the much easier seed phrase login?

I went with II just because I had already recently figured out how to do that locally anyway so it was fresh.
The tricky part was the transfer of tokens anyway. I got completely caught by the e8s conversion 😆 The II stuff worked on the first try, somehow...

> Is the II canister installed with the rest of NNS canisters?

Yes it is! It's not documented for some reason so I was originally installing II manually, but I spotted it in the install logs for the NNS canisters!

src/frontend/src/header.jsx
nathanosdev: If the ID goes inside `CopyToClipboard` then it would need to be a generic ID for all instances of `CopyToClipboard`, but I needed unique IDs for those cases and in general prefer to keep them unique, instead of querying for all instances and then accessing by array index.

For `BurgerButton`, it could definitely go inside. I felt that the `Header` component was more likely to be only used once and `Burger` _could_ be used again somewhere else, although it's probably unlikely.

nathanosdev: Or do you mean to do it like this?
`<BurgerButton data-testid="burger-button" />`

nathanosdev: Yeah that makes sense. I wasn't sure about actually changing components to make them work for e2e tests, but if you don't mind having to account for that in the components props then it has it's advantages over creating unnecessary wrapper elements. I'll go with that approach.

nathanosdev:
@taggrx I've updated the components to take `testId` as a prop, like we discussed and I've disabled the login with II test for now. If we merge this then I can work on the login with seed phrase test and build on the common stuff that's part of this PR.

chmllr (approved):

taggrx:
Thanks so much Nathan! 🙏 This looks amazing and I can't wait to have the first e2e tests as we saw more than once lately how important they are!

taggrx (approved):

PR #86, opened 2023-04-22
Lazy distributions

taggrx:
The current distribution process computes the amount of e8s per user and then makes an individual transfer from the treasury to each user. Since every transfer involves a call to the ICP ledger, it has two big disadvantages:

- the distribution is very slow and lasts many minutes,
- it is costly for the Taggr canister.

Instead my proposal is to simply record the e8s owed to the user by the treasury and transfer all of them whenever the user mints cycles or executes any other ICP transaction from their Taggr wallet. The number of user's ICP in the treasury will be also displayed on user's wallet.

This MR also removes the deprecated code related to seed phrase v1 from the wallet.


nathanosdev:
Code LGTM!

But, and please excuse my ignorance of canister development (I'm working on it 😂), if we assume that everyone that gains from the distribution takes some action to get their ICP before the next distribution, will the end result of cycles consumed by the Taggr canister not be the same since the same number of transactions will have been executed? Does running this alongside another transaction somehow reduce the number of cycles consumed by the canister?

taggrx:
> Code LGTM!
> 
> But, and please excuse my ignorance of canister development (I'm working on it 😂), if we assume that everyone that gains from the distribution takes some action to get their ICP before the next distribution, will the end result of cycles consumed by the Taggr canister not be the same since the same number of transactions will have been executed? Does running this alongside another transaction somehow reduce the number of cycles consumed by the canister?

Great question! My assumptions are:

1. It is unlikely that everyone will withdraw from the treasury after every distribution, but I indeed judge by myself. I mint cycles less than once per week and almost never transfer the ICP from Taggr.
2. If the assumption 1 fails we indeed do not save cycles, but we at least turn the distribution into a quick synchronous operation. Once have e.g. 10k users and each transfer takes about 2 seconds, then a distribution will take 5 hours! It just does not scale.

nathanosdev:
> > Code LGTM!
> > But, and please excuse my ignorance of canister development (I'm working on it 😂), if we assume that everyone that gains from the distribution takes some action to get their ICP before the next distribution, will the end result of cycles consumed by the Taggr canister not be the same since the same number of transactions will have been executed? Does running this alongside another transaction somehow reduce the number of cycles consumed by the canister?
> 
> Great question! My assumptions are:
> 
> 1. It is unlikely that everyone will withdraw from the treasury after every distribution, but I indeed judge by myself. I mint cycles less than once per week and almost never transfer the ICP from Taggr.
> 2. If the assumption 1 fails we indeed do not save cycles, but we at least turn the distribution into a quick synchronous operation. Once have e.g. 10k users and each transfer takes about 2 seconds, then a distribution will take 5 hours! It just does not scale.

Thanks for the explanation! That makes a lot of sense, especially with regard to scaling.

If people complain, maybe we could consider creating an option to manually withdraw from treasury to wallet. If someone needs more cycles and doesn't have enough ICP in their wallet but they have enough in the treasury then they'll have to transfer more ICP to be able to mint.

nathanosdev (dismissed):

taggrx:
>  If someone needs more cycles and doesn't have enough ICP in their wallet but they have enough in the treasury then they'll have to transfer more ICP to be able to mint.

But this is already in the PR: if you mint cycles and your treasury balance is > 0, first this ICP is moved to your wallet and then the cycles are minted from that. So it does not even create any new frictions for the user. The only difference is that instead of having everything on the ICP account, the user sees 2 balances (ICP acc + treasury), but they behave as a single one. The reason why I decided to not simply display the sum of both is because I do not want to show a different ICP balance on user's account than they would see on the IC dashboard.

radu-daniel (dismissed):

mechaquan:
Cool idea!  Maybe the "owed" ICP should be transferred to a different "treasury"...

I'd also suggest allowing the users to make their own transfer and pay the tx fee.

For example, let's say some accumulates several ICP they can then initiate a transfer minus 0.0001 ICP.

nathanosdev:
> > If someone needs more cycles and doesn't have enough ICP in their wallet but they have enough in the treasury then they'll have to transfer more ICP to be able to mint.
> 
> But this is already in the PR: if you mint cycles and your treasury balance is > 0, first this ICP is moved to your wallet and then the cycles are minted from that. So it does not even create any new frictions for the user. The only difference is that instead of having everything on the ICP account, the user sees 2 balances (ICP acc + treasury), but they behave as a single one. The reason why I decided to not simply display the sum of both is because I do not want to show a different ICP balance on user's account than they would see on the IC dashboard.

Ah yes, I read that but it didn't click 🤣 
Makes sense, thank you!

src/backend/env/mod.rs
```diff
@@ -976,28 +979,17 @@ impl State {
...
-                        e8s_to_icp(user_reward),
-                        e8s_to_icp(user_revenue)
-                    ));
-                }
-                Err(err) => {
-                    self.logger
-                        .error(format!("Couldn't transfer ICP to @{}: {}", &user.name, err));
-                }
-            }
+            user.unclaimed_e8s += e8s;
```
chmllr, line 982: Maybe a bit more straightforward name, like `treasury_e8s`? (but it's a tiny nit)

chmllr (approved):
LGTM

src/backend/env/mod.rs
taggrx: Will change, thanks!

taggrx:
> Can you see my comment on the review commit?
> 
> It says threasury... is this a typo?

Oops 🙈 Will fix, thanks!

PR #89, opened 2023-04-24
fix post timestamp computation

taggrx:
We currently have a minor bug on how the post's timestamp is computed which leads to delay on polls: whenever a poll is edited, the delta to the original and the original timestamp are moved to the `patches` vector. Hence, to compute posts real timestamp, one needs to take the minimum from all versions.


nathanosdev (approved):

nathanosdev:
Changes make sense 👍

> "the delta to the original and the original timestamp are moved to the patches vector"

I'm pretty sure I understand this, but just to make sure. Does this mean that the post itself will always have the most recent timestamp as it's timestamp? So whenever you edit the post, the current version is pushed into the patches vector?

src/backend/env/post.rs
```diff
@@ -81,6 +81,17 @@ impl Post {
         }
     }
 
+    // Return post's original timestamp, either by taking the minimum from the edits or the one
+    // assigned to `timestamp`.
+    pub fn timestamp(&self) -> u64 {
```
nathanosdev, line 86: If the original timestamp is important, do you think it could justify it's own property that never changes? To save on calculating it each time?

taggrx: Well the calculation is negligible here but if it ever would turn out a bottle neck - sure!

PR #90, opened 2023-04-25
Better karma leaders

taggrx:
Reduces weekly karma leader to 12 spots and displays in the same style as all dashboard values.

Screenshot 2023-04-25 at 16 00 29


mechaquan (approved):
Is changing to 12 for aesthetics only?  Will there be a way to see more? ... if I was karma farming I would just make sure my alt is 13th! :)

nathanosdev:
I agree this looks better and the code looks good so I'll approve 👍 
I'm not too opinionated about how many karma leaders are shown, so maybe this is a question for the DAO and @mechaquan has a good point about providing a way to see more. This is probably not "karma leaders" anymore, but just a separate summary page of all karma changes in the current cycle.

nathanosdev (approved):

taggrx:
> Is changing to 12 for aesthetics only? Will there be a way to see more? ... if I was karma farming I would just make sure my alt is 13th! :)

But before it was also a fixed number 24, so before you would try to be 25th...? 😂 I see what you mean of course, but if you farm you can easily just try to not be in top 5 to attract attention. And some farming "on a back burner" is even impossible to tackle, right? I'd try to target the farming systemically. E.g the karma sources introduced recently that we hopefully finally can deploy at some point is one much more constructive step into this direction... And I think we need more of that.

taggrx:
> I agree this looks better and the code looks good so I'll approve 👍 I'm not too opinionated about how many karma leaders are shown, so maybe this is a question for the DAO and @mechaquan has a good point about providing a way to see more. This is probably not "karma leaders" anymore, but just a separate summary page of all karma changes in the current cycle.

Well, they are still sorted as before, just reduced to 12 instead of 24 and displayed uniformly with other dashboard infos. So I'd argue these people are still weekly karma leaders.

@mechaquan if you think e.g. it's useful to have a full list of karma people, I can work on that. But I also plan to create full reports after each distributions because what we have will not work with 1k of users, e.g. Once we have this feature, investigating these reports will probably provide enough infos for identifying farmers, wdyt?

mechaquan:
> @mechaquan if you think e.g. it's useful to have a full list of karma people, I can work on that. But I also plan to create full reports after each distributions because what we have will not work with 1k of users, e.g. Once we have this feature, investigating these reports will probably provide enough infos for identifying farmers, wdyt?
> 
> 
> 
> 

Yeah I agree we don't want to just expand for eternity... I feel like the optimal approach is a aesthetically pleasing display and then a link to a separate page with filters/search or something.

I don't think it needs to be part of the main menu or anything just a link on the dashboard.

nathanosdev:
> So I'd argue these people are still weekly karma leaders.

Sorry, I meant that the new page showing all users with their karma for the current cycle would not be karma leaders. So I suggested a separate page, rather than extending the dashboard with an option to show all karma.

chmllr (approved):

PR #91, opened 2023-04-26
reimplement reposts as extensions

taggrx:
The current reposting functionality is very old and is actually a hack: the markdown parser finds a string of a form `[repost:1234](#/post/1234)` and renders the corresponding post in-place. This has a lot of drawbacks like inability of the backend to recognize reposts, rendering issues in previews and so on. This PR re-implements reposts as post extensions (like polls or proposals).

The regex dependency will be deleted again after this change is deployed and all posts are migrated.


nathanosdev (approved):
LGTM 👍

chmllr (approved):

PR #93, opened 2023-05-01
Always use local version of Playwright and do not run the HTML server when tests fail

nathanosdev:
This PR fixes an issue where the global Playwright (installed with `npx playwright`) is installing a different Chromium version than what the local Playwright (installed with `npm install`) wants, this was causing tests to fail. It also adds a configuration that prevents the HTML report server from running when tests fail, this was causing the Docker container to never exit when tests failed.

See the thread on Taggr for details: https://taggr.link/#/post/25317


Dockerfile
```diff
@@ -39,8 +39,12 @@ RUN mkdir -p /opt/ic-wasm && \
 # Install dfx
 RUN sh -ci "$(curl -fsSL https://internetcomputer.org/install.sh)"
 
+# Install NPM dependencies
+COPY package.json package-lock.json ./
```
nathanosdev, line 43: By copying only `package.json` and `package-lock.json` we make sure the `npm ci` and `npm run install:e2e` only run again when those files change, so the Docker layer caching is more efficient. If we put `COPY . .` here then those two commands would run again whenever any file changes.

This won't affect the CI pipeline, but will make the Docker build faster for local development or build hash reproduction.

taggrx (approved):
Great! Thanks so much Nathan for the quick fix! 🙏

PR #95, opened 2023-05-03
Minor changes and refactorings

taggrx:
No real user facing changes, only small internal improvements.


src/backend/env/config.rs
```diff
@@ -190,7 +190,7 @@ pub const CONFIG: &Config = &Config {
     min_stalwart_account_age_weeks: 26,
     stalwart_moderation_reward: 20,
 
-    hot_post_engagement_percentage: 0.008,
+    hot_post_engagement_percentage: 0.01,
```
nathanosdev, line 193: I don't know this too well yet, but this will make it easier for posts to show up in the "hot" feed?

src/backend/env/mod.rs
```diff
@@ -1621,42 +1636,25 @@ impl State {
             users: self.users.len(),
             posts,
             comments: self.posts.len() - posts,
-            cycles,
+            cycles: self.users.values().map(|u| u.cycles()).sum(),
```
nathanosdev, line 1639: Why not also calculate this sum in the main `self.users.values()` loop above?

src/frontend/src/proposals.jsx
```diff
@@ -97,7 +97,7 @@ export const Proposal = ({id}) => {
         const newStatus = data.status;
         if (prevStatus == "Open" && newStatus == "Executed" && "Release" in data.payload) {
             setStatus("Executing the upgrade...");
-            await api.call("execute_upgrade", true);
+            await api.call("execute_upgrade", false);
```
nathanosdev, line 100: It looks like this argument is not used on that method: https://github.com/TaggrNetwork/taggr/blob/main/src/backend/lib.rs#L175-L190

I also don't see it used on the client side: https://github.com/TaggrNetwork/taggr/blob/main/src/frontend/src/api.js#L39-L44

Am I missing something?

src/backend/env/config.rs
taggrx: This is just a percentage of engagements (reactions + comments) wrt. total user number. This change increases the number to roughly 11 from 8 for a post to be "hot". So no, it makes it more difficult to stand out.

src/backend/env/mod.rs
taggrx: Oh I missed that one, thanks so much!

src/frontend/src/proposals.jsx
taggrx: `execute_upgrade` is a JS method, it does not appear in `api.js` explicitly. But it is used: https://github.com/TaggrNetwork/taggr/blob/main/src/backend/lib.rs#L186

The purpose of this flag is to not upgrade if there are outstanding canister calls. There was a problem in the past, that some calls were "never closed" because the continuation after the `await` has panicked. But this is solved now and this is why I'd like to take the calls into consideration again. In the worst case, this call can be repeated with a `true` argument from the JS console.

nathanosdev: Ah I missed this `arg_data_raw` bit, I was expecting the parameters in the method. Okay makes sense, thanks!

nathanosdev (approved):

taggrx:
@nathanosdev sorry I didn't see you approved already and just pushed another set of UI fixes I discovered locally: https://github.com/TaggrNetwork/taggr/pull/95/commits/5ff36499269aada9238f66a9725510909d5e8b83 (the biggest chunk is fixing indentation, not the code itself.)

nathanosdev (approved):

chmllr (approved):

PR #96, opened 2023-05-04
Fixes

taggrx:
Fixes preview issues and the registration. Urgent.


src/frontend/src/form.jsx
```diff
@@ -34,6 +34,11 @@ export const Form = ({postId = null, comment, realmArg = "", expanded, submitCal
             alert("Your post does not fit on screen without scrolling.\n\nPlease add a cut line (three empty lines) after the introductory part.");
             return false;
         }
+        let parts = value.trim().split(CUT);
+        if (parts.length != 2 || parts[0].length > parts[1].length) {
```
chmllr, line 38: I think this will break posts without cuts. You need to combine it with the check above.

chmllr (approved):

src/frontend/src/form.jsx
taggrx: Excellent catch! 🙏

chmllr (approved):

PR #97, opened 2023-05-04
Http request update

taggrx:
Upgrades the query to the update call if no certified asset was found. Since this is a critical change, I tested it by deploying a new canister https://e4i5g-biaaa-aaaao-ai7ja-cai.icp0.io. Now, we share URLs without hashes and they return an `index.html` corresponding to the shared resource. E.g. visit https://e4i5g-biaaa-aaaao-ai7ja-cai.icp0.io/user/xx in the browser to make sure the link works and test `curl https://e4i5g-biaaa-aaaao-ai7ja-cai.icp0.io/user/xx` in the console to make sure the served page contains all headers.

Thanks so much @chmllr and @nathanosdev for making this happen!


chmllr (approved):
Nice! 🚀

PR #99, opened 2023-05-05
Object manager simplifications

taggrx:
In this MR I simplify the object manager component to a simple read/write stable memory based store for arbitrary objects. With that it will be used soon to move old posts to the stable memory to avoid running into the instruction limit on an upgrade.

All these changes do not affect any production code yet.


src/backend/env/memory.rs
```diff
@@ -266,115 +245,29 @@ impl Allocator {
...
-    2000
 }
 
-impl<K: Eq + Ord + Clone + Display, T: Storable> ObjectManager<K, T> {
-    pub fn keys(&self) -> impl Iterator<Item = &'_ K> {
-        self.index.keys()
-    }
-
-    pub fn insert(&mut self, id: K, value: T) -> Result<(), String> {
+impl<K: Eq + Ord + Clone + Display> ObjectManager<K> {
```
chmllr, line 254: High level tests for the object manager would make sense.

chmllr (approved):

src/backend/env/memory.rs
taggrx: Good idea, thanks! I'll add them once the object manager is actually used in the mainnet code.

PR #100, opened 2023-05-06
Fixes

taggrx:
This PR fixes two bugs.

1. The current implementation of $TAGGR transfers from the UI is using an internal transfer function which is _not checking_ if the principal is the minting account. But since the minting account right now is set to the anonymous principal (we might want to change this as well in the future), what could be possible if the transfers were enabled is that anybody could make sure they're not logged in into Taggr and then initiate a "transfer" for any token amount from the JS console. This would effectively result in a valid minting transaction.
2. The double comment bug created when editing a post is fixed now too.


nathanosdev:
Are there any `update` methods that are valid when called from an anonymous principal?

nathanosdev (approved):

chmllr (approved):

taggrx:
> Are there any `update` methods that are valid when called from an anonymous principal?

Yes, for example `finalize_upgrade` and `execute_upgrade` because they work only once after an upgrade proposal is adopted and it doesn't matter who triggers them.

PR #101, opened 2023-05-07
Refactors the funding proposal ui

taggrx:
Reimplements the funding proposal UI from JS-prompt based to an HTML based mask.


nathanosdev:
I'm testing this out locally, funding form looks good 👍 much better than using JS prompts.
But I found it a little strange to be able to show both forms at once, is it on purpose?

I think it might be better to have this work a little bit like tabs, so maybe something more like this:
```
const [currentMask, setCurrentMask] = React.useState('release');

// ...

<button onClick={() => setCurrentMask('release')}>RELEASE</button>
<button onClick={() => setCurrentMask('funding')}>FUNDING</button>

// ...

{currentMask === 'release' && // ... release form}
{currentMask === 'funding' && // ... funding form}
```

It might be nice to also have the `active` class switch between the RELEASE and FUNDING buttons so it's clear which form is currently active.

nathanosdev (approved):

PR #102, opened 2023-05-07
add login with seed phrase e2e test

nathanosdev:
This PR adds an e2e test for login with a seed phrase.
 
I also added a new `optimize` option that's available from `ic-wasm` now: https://github.com/dfinity/ic-wasm/#shrink
This made a _tiny_ difference in filesize, less than 10kb so that's not even worth talking about, but apparently, it reduces cycle consumption by 4%, if that turns out to be true then that would be awesome.


package.json
```diff
@@ -20,6 +20,7 @@
     "@dfinity/principal": "0.15.5",
     "@playwright/test": "^1.32.3",
     "@types/isomorphic-fetch": "^0.0.36",
+    "bip39": "^3.1.0",
```
taggrx, line 23: But the seed phrase doesn't have to be a bip39 phrase, it can be an arbitrary string. If there are no strong arguments for bip39, I'd either use a hardcoded bip39 seed phrase or just any string instead of adding dependencies (even if dev dependencies). WDYT?

nathanosdev: I see, I'll use a random generated string in that case. A hard coded phrase will restrict us in tests that need multiple users.

e2e/tests/sign_up.spec.ts
```diff
@@ -44,3 +48,39 @@ test.skip("sign up with internet identity", async ({ page, baseURL }) => {
   await expect(homePage.connectButton).not.toBeVisible();
   await expect(authenticatedHomePage.burgerButton).toBeVisible();
 });
+
+test("sign up with seed phrase", async ({ page, baseURL }) => {
```
taggrx, line 52: Awesome! 🙌

taggrx (approved):

taggrx:
Thanks so much Nathan!

PR #110, opened 2023-05-13
Use the decentralized neuron

chmllr:
This switches Taggr to the new decentralized neuron and adds an `update` method `get_neuron_info` which is proving, that Taggr is able to obtain the full neuron info. It also prints the controller, the followees on topic 1 and the hot keys (which is the Taggr canister).

@taggrx I am not sure how to add the UI part, could you please help with this?


taggrx:
Thanks Christian! I'll add a WP section now!

nathanosdev (approved):

PR #112, opened 2023-05-13
add simple post creation e2e tests

nathanosdev:
This PR adds e2e tests for simple text-based post creation and posts with hashtags, also testing that the hashtag feed works.

I've removed the signup tests since signup is now implicitly tested by the post-creation tests so the signup tests would be redundant. The code to sign up users has been moved to reusable functions so we can easily use that in any test.

A note on Playwright's terminology, a browser "context" basically refers to an incognito window. So when some code is run in a "new context" it has a completely fresh state in the browser. I've used this to create posts from different users in the same test.


taggrx (approved):
I struggle to understand many things in detail (or rather the APIs), but I can hopefully catch up once I start writing my own tests.

Thanks a lot Nathan!

PS: How come you're so proficient with this framework? Are you a frontend developer?

nathanosdev:
> I struggle to understand many things in detail

Please do let me know if there's anything I can help clarify, but yes I think things will be much clearer once you get your hands dirty with it yourself.

> How come you're so proficient with this framework? Are you a frontend developer?

Yes, that's where most of my experience was prior to joining Dfinity :)

PR #113, opened 2023-05-15
Improvements & urgent bug fix

taggrx:
This fixes the image uploading bug which was introduced by a new icon `Image` which is shadowing the JS class :(


src/frontend/src/header.jsx
```diff
@@ -3,17 +3,26 @@ import {BurgerButton, ButtonWithLoading, HeadBar, Loading, ReactionToggleButton,
...
-    React.useEffect(() => { document.getElementById("logo").innerHTML = backendCache.config.logo }, []);
-    React.useEffect(() => { toggleButtonBar(false); toggleRealms(false) }, [route]);
+    React.useEffect(() => { document.getElementById("logo").innerHTML = backendCache.config.logo; }, []);
+    React.useEffect(() => {
+        toggleButtonBar(false);
+        toggleRealms(false);
+        if (route.includes("inbox"))
+            interval = setInterval(() => setMessages(Object.keys(user.inbox).length), 1000);
+        else clearInterval(interval);
+    }, [route]);
```
nathanosdev, line 25: Can this be simplified by introducing `user` as a dependency for `effect` so it runs when `user` changes? This would avoid the need to "poll" the user object.

nathanosdev (approved):

src/frontend/src/header.jsx
taggrx: I don't think it is possible with the current architecture. The `user` is a plain JS object anchored in the `window` space. There is no hook, which would trigger the re-rendering of the `Header` React component if this object was changed. And if `Header` is not re-rendered, no `effect` hooks would trigger.

I think such issues are usually solved by another JS framework managing the app state (I forgot the name), but I never learned it and this is in fact the first place in Taggr, where I need a _common_ state into two react components which are rendered independently (maybe this should be changed first?).

nathanosdev: Got it! Yes I think using a state management framework would help with stuff like this. The name you're thinking of is probably Redux? that's the most popular one.

I've been thinking about introducing that as a next step after the e2e tests. If you're interested, I've been using this one recently and really like it because it fits into React's hook ecosystem really nicely and it has way less boilerplate nonsense than the other solutions: https://github.com/pmndrs/zustand

nathanosdev (approved):

src/frontend/src/header.jsx
taggrx: Yes Redux. I had a chance to work on a project using it in the past, but as you say - the amount of boilerplate and all the indirections made me question its benefits in the end. 

Regarding `zustand`, I'd be very interested if it helps us to simplify the code. On the first glance I like it.

PR #116, opened 2023-05-16
post creation with image e2e test

nathanosdev:
This PR adds an e2e test for image uploading on posts. 3 days too late to prevent a bug but at least we have it now 🤣


e2e/tests/post_creation.spec.ts-snapshots/uploaded-img-chromium-linux.png
nathanosdev, line 1: This file is a screenshot of the page after clicking the image in the post. This was the easiest way I could find to assert that the expected image is uploaded properly. On subsequent runs of the test, Playwright will take a new screenshot and compare it to this one. If the comparison fails, the test fails. If for some reason, we expect this to change then it can be manually updated (command is added to the docs).

e2e/tests/post_creation.spec.ts-snapshots/post-creation-with-image-1-chromium-linux.png
nathanosdev, line 1: This file is a screenshot of the page after clicking the image in the post. This was the easiest way I could find to assert that the expected image is uploaded properly. On subsequent runs of the test, Playwright will take a new screenshot (but won't save it) and then compare it to this one. If the comparison fails, the test fails. If for some reason, we expect this to change then it can be manually updated (command is added to the docs).

e2e/tests/post_creation.spec.ts
```diff
@@ -60,3 +62,25 @@ test("post creation with hashtag", async ({ page, baseURL, browser }) => {
...
+
+  const postPage = await newPostPage.submit();
+  const uploadedImage = postPage.postBody.locator("img");
+  await expect(uploadedImage).toBeVisible();
+
+  await expect(postPage.imagePreview).not.toBeVisible();
+
+  await uploadedImage.click();
+  await expect(postPage.imagePreview).toBeVisible();
+  await expect(postPage.imagePreview.locator("img")).toHaveScreenshot();
```
nathanosdev, line 82: This takes a screenshot of the page and saves it to a file for future comparisons

e2e/tests/post_creation.spec.ts-snapshots/post-creation-with-image-1-chromium-linux.png
taggrx: Is the comparison on a bit level? If yes, why not just taking an md5 hash?

nathanosdev: > Is the comparison on a bit level? 

Yes, although it's a bit more complex than that. It supports some degree of variance in the comparison which can be configured, which can be useful if you're okay with an element moving 5-10 pixels on the page, but not jumping 500 pixels. In this case, it's not relevant, but it might be relevant in screenshots from future tests, especially if we start testing designs across multiple browsers and comparing screenshots from each browser.

> why not just taking an md5 hash?

I think that would be incredibly difficult to debug when it goes wrong. Debugging two differing screenshots is much quicker than debugging two differing hashes.

And how do you determine whether the first hash is correct or not? When it's a screenshot, we can manually check that the first screenshot looks as it is intended to look, which is a very quick operation and it's something that can be easily done as part of the pull request. And then the framework automates checking that subsequent runs do not deviate from the original one.

If it's an MD5 hash, how do we know what the MD5 hash should be? We would need to manually run the sample image through the resizing logic (or any other future transforms we might have for images) to find out the hash of the uploaded asset, then download the image from the page, hash it, and compare it to our manually calculated hash. This is something that also should be done by code reviewers to ensure that it's done correctly. I think it introduces a lot more complexity.

chmllr (approved):

e2e/tests/post_creation.spec.ts-snapshots/post-creation-with-image-1-chromium-linux.png
taggrx: Thanks a lot Nathan, I understand now!

taggrx (approved):

PR #120, opened 2023-05-18
post interaction e2e tests

nathanosdev:
This PR adds a test where a post is reacted to with a heart. It also adds cycle and karma checks to the existing post creation tests and fixes the issue with the post creation with images.


e2e/pages/new_post_page.ts
```diff
@@ -32,15 +32,19 @@ export class NewPostPage {
...
 
   public async submit(): Promise<PostPage> {
     await this.sendButton.click();
 
     // since this navigation is asynchronous and not a result of directly
-    // clicking an anchor tag, Playwright does not know that it needs to wait
-    await this.page.waitForURL("**/#/post/*");
+    // clicking an anchor tag, Playwright does not know that it needs to wait.
+    // wait for network idle so we know any post images are finished loading
+    await this.page.waitForURL("/#/post/*", { waitUntil: "networkidle" });
```
nathanosdev, line 47: This is the fix for the post creation with image failure. The post images were not always loading before the tests were executed. Waiting for network requests to finish first fixes this.

e2e/support/context.ts
```diff
@@ -1,15 +1,12 @@
...
 
 export async function performInNewContext<T>(
   browser: Browser,
   task: (page: Page) => Promise<T>
-): Promise<T> {
+): Promise<[Context, T]> {
   const context = await browser.newContext();
   const page = await context.newPage();
 
-  try {
```
nathanosdev, line 10: Calling `context.close()` here was causing Playwright's UI mode to crash, even if there was nothing wrong with the test. This was making debugging tests harder. To fix this, we need to return the context and manually close it at the end of the test.

src/backend/lib.rs
```diff
@@ -712,7 +712,15 @@ fn user() {
                 .flat_map(|post| post.reactions.iter())
                 .flat_map(|(r_id, users)| {
                     let cost = karma.get(r_id).copied().unwrap_or_default();
-                    users.iter().map(move |user_id| (*user_id, cost))
+                    users
```
nathanosdev, line 715: This fixes a very minor issue I noticed where a bootcamp user that only has a reaction from another bootcamp user will have 100% of their karma on the "karma from last 90 posts" section of their profile, which is a harmless issue but can be confusing since they haven't actually received any karma from them.

src/frontend/src/index.jsx
```diff
@@ -41,33 +41,7 @@ const parseHash = () => {
...
-    } else if (currentFrame && currentFrame.dataset.hash == location.hash) {
-        currentFrame.remove()
-    }
-
-    let frame = document.createElement("div");
-    frame.dataset.hash = location.hash;
-    stack.appendChild(frame);
-    createRoot(frame).render(content);
-}
+const root = createRoot(document.getElementById('root'));
```
nathanosdev, line 44: I didn't realize the navigation was working like this until it was causing issues with the e2e tests, since there would be multiple instances of the same element in the HTML. It wasn't picked up before because the e2e tests were not navigating back and forth from the same page yet. Now they are with the profile page.

Removing this hasn't caused any issues that I can see with my manual testing, the e2e testing still works. I think it's better to not have this because it increases the size of the DOM every time a page navigates.

@taggrx it would be great if you could clarify if there are any edge cases I might be missing here and what the motivation behind this was.

taggrx: This is what makes the app behave like a native app on mobile: when you go back to the previous page it does not reload and does not render from scratch, but shows you the same view with even the same scrolling position. We should not delete this. Maybe we can disable it for e2e tests? Or even better - is there a proper way of solving this issue?

nathanosdev: Ah I see, thanks for clarifying what the motivation for this was.

I would prefer to avoid disabling/enabling different code for e2e tests as much as we can so we can be sure that we don't get any false positives/negatives in the tests, but an easy work around for now would be to use the `:visible` pseudoselector to filter what elements we select.

Alternatively, have you considered using a library for this? https://reactrouter.com/en/main

Or, if there's strong reasons not to use a library, I think we can achieve the scroll position restoration and avoiding the page reloads with a custom `Link` component.

For the re-rendering, if you set `display: none` and then `display: block` the browser will re-render those DOM elements regardless of what we do. If it's about avoiding some initialization logic in the components, then maybe the answer here is to move that logic into a state management library like we briefly discussed on a different PR. With the state separated (and memoized) from components, bootstrapping a component from scratch every time has a much lower cost. I think this is much better than maintaining a history in the DOM because that can consume a lot of memory if someone is spending a lot of time on Taggr.

taggrx: The reasons why I try to avoid libraries:

- currently the entire fronted is served as a single file which is already too large and is a big problem if you believe various online tools analyzing the web site performance
- the Wasm size needs to stay below 2 MB
- Security of course (the more dependencies the more ways there are to compromise Taggr)

After I have built this caching, it worked surprisingly well and never caused any problems yet (it only grows significantly if you indeed spend weeks without reloading).

But of course I am not against _any_ improvements especially when done right. Maybe we should fix it right now with the `visible` selector and eventually revisit as a separate task?

nathanosdev: I think this makes sense and I would also avoid libraries for the same reasons. But I do think that having a small selection of good libraries can go a long way in improving the overall quality of the codebase. I usually have at least router, state management and a CSS solution (which may also just be Sass, so that's not a runtime dependency). The router in this case can actually help us reduce the size of the codebase by introducing the ability to use lazy loading and splitting that file into smaller pieces.

And yes I agree with going with the visible selector for now and then revisiting this.

taggrx:
This one cannot be reviewed yet, right? I think this `:visible` fix didn't go in yet, right?

nathanosdev:
> This one cannot be reviewed yet, right? I think this :visible fix didn't go in yet, right?

Yes I'll need to merge the other PR first and pull it into here to make sure everything is still okay with these tests and then make the `:visible` fix too. I've converted it back into a draft PR for now and I'll mark it as ready once I've finished it.

nathanosdev:
@taggrx this is ready for review now :)

src/frontend/src/post.jsx
```diff
@@ -331,11 +331,27 @@ export const Reactions = ({reactionsMap, react}) => {
...
+                onClick={() => react(reactId)}  data-testid={reactionToTestId(reactId) + "-reaction"}>
                 {reaction2icon(reactId)}&nbsp;{`${users.length}`}
             </button>;
         })}
     </div>;
 };
 
 const skipClicks = elem => elem && (elem.dataset["meta"] == "skipClicks" || skipClicks(elem.parentElement));
+
+const reactionToTestId = id => {
```
taggrx, line 343: Btw, wouldn't it make more sense to use the id numbers as is, but define the mapping for generated ids (`<n>-reaction -> <wrods>`) in the testing code?

taggrx (approved):

taggrx:
Thanks so much Nathan!

src/frontend/src/post.jsx
nathanosdev: Yes, I like this. I'll update it to use this approach, thank you!

PR #121, opened 2023-05-19
Fixes

taggrx:
This PR contains 5 fixes:

- fixes the missing cut line in comment previews,
- refunds tips on post deletions and removes tags,
- decouples the upgrade proposal execution and the actual upgrade execution: now the hourly routine will check for pending upgrades and execute one if it finds one (context: https://taggr.link/post/27791). This will now happen _after_ Taggr checked for NNS proposals which will dramatically minimize the probability of running into open canister calls,
- fixes notification counter refresh,
- the watched post notifications will now only display the post id which is the tail of a thread, instead of all new messages on the whatched post: this the user an catch up much quicker by simply covering all thread bottom up instead of clicking on each new message returning to the same thread.

It also move thee nns related logic into a separate function.


src/backend/env/canisters.rs
```diff
@@ -164,10 +164,12 @@ pub fn upgrade_main_canister(logger: &mut Logger, wasm_module: &[u8], force: boo
     logger.info("Executing the canister upgrade...");
     let calls = calls_open();
     if calls > 0 && !force {
+        let calls_map = unsafe { &CALLS };
         logger.error(format!(
-                "Upgrade execution failed: {} canister calls are in-flight: {:?}. Please re-trigger the upgrade finalization.",
-                calls,
-                unsafe { &CALLS }
+            "Upgrade execution failed due to open canister calls: {:?}",
+            calls_map
```
chmllr, line 170: Consider using `if let Some(calls_map) = unsafe { &CALLS } { ...` instead to avoid `Some` in the logs.

chmllr (approved):

chmllr:
Can't wait for the notifications fixes 😅

PR #124, opened 2023-05-22
post creation with images e2e test (again)

nathanosdev:
This reverts the reverted commit with the post creation e2e tests and also fixes the dockerfile's entry point so it exits with a non-zero exit code when the tests fail. I haven't fixed the tests yet because I want to prove it works :)


taggrx:
Yay, it works! :D

nathanosdev:
I'm having issues with the image loading inside the docker container. It works fine when it's run outside of the Docker container. It's not an issue of the image taking too long to load. I've honestly no idea what's happening.

@taggrx @chmllr how do you guys feel about just running the tests directly in a GitHub action? It won't block the `make release` command anymore, but it will at least block PRs from being merged.

Or if either of you knows how to debug a browser running inside a Docker container, then maybe this might give us some clue as to what's going on.

taggrx:
> @taggrx @chmllr how do you guys feel about just running the tests directly in a GitHub action? It won't block the `make release` command anymore, but it will at least block PRs from being merged.

Sure this would prevent having a green checkmark on commits, right? I think this is all we need. Maybe we can even make `make test` run `make e2e_test` and make it a separate github action instead of part of release building?

taggrx:
Oh wait it looks like `make test` is not even executed anymore during `make release`?! 😱

taggrx:
> I think making it part of make test makes sense for running everything locally in one go, but for the GitHub pipeline we can go one step further and run the e2e tests in parallel with the unit tests and linting.

Yes this is true, let's do that!

Do you want to make it in this MR or should I already add the `make test` back to `make release`?

taggrx:
Ok, great! Thanks a lot! And don't hurry, I don't plan to create new releases in the next days, so this is not urgent.

nathanosdev:
Superseeded by https://github.com/TaggrNetwork/taggr/pull/128

PR #125, opened 2023-05-22
fix poll radio groups

nathanosdev:
This PR fixes an issue with the poll radio inputs reported here: https://taggr.link/post/28185


src/frontend/src/poll.jsx
```diff
@@ -21,9 +22,10 @@ export const Poll = ({poll, post_id, created}) => {
             const votes = (data.votes[id] || []).length;
             const pc = totalVotes > 0 ? Math.ceil(votes / totalVotes * 100) : 0;
             return <label key={id} className={showVoting ? "vcentered" : null} style={{display: "flex", flexDirection: showVoting ? "row" : "column"}}>
-                {showVoting && <input type="radio" value={id} name={id} className="right_spaced" style={{marginTop: 0}}
+                {showVoting && <input type="radio" value={id} name={radio_group_name} className="right_spaced" style={{marginTop: 0}}
```
nathanosdev, line 25: A radio input group should have a consistent `name` across each option. Having a different `name` means the browser will recognize them as different groups, which is why it was possible to select all options.

The core fix is actually the `e.preventDefault()`, which stops the browser from showing the radio option as selected when it's clicked. So having a consistent `name` isn't necessary, but we should have it anyway for accessibility.

We don't actually need to display the input as selected because as soon as it is selected it is replaced by the poll results.

taggrx (approved):
Awesome! Thanks so much!

PR #126, opened 2023-05-24
Rewards aka weighted funding

taggrx:


src/backend/env/config.rs
```diff
@@ -219,7 +219,7 @@ pub const CONFIG: &Config = &Config {
 
     reaction_fee: 1,
 
-    max_funding_amount: 10000,
+    max_funding_amount: 20_000_00, // at ratio 1:1
```
nathanosdev, line 222: Is this measured in whole Taggr tokens? 2 million seems way too high, or am I missing something?

Minor thing, but I'd find it easier to read as `2_000_000`.

src/backend/env/proposals.rs
```diff
@@ -44,6 +76,33 @@ impl Proposal {
             .get(&account(principal))
             .ok_or_else(|| "only token holders can vote".to_string())?;
 
+        match &mut self.payload {
+            Payload::Release(release) => {
+                if release.hash != data {
+                    return Err("wrong hash".into());
```
nathanosdev, line 82: Nice! I had this on my todo list 😄

src/backend/env/config.rs
nathanosdev: Ah, after reading further in `proposals.rs` this makes more sense now.

nathanosdev (approved):

src/backend/env/config.rs
taggrx: Yep, this is why I put the `_` before the last 2 zeros, to kind of make clear those are the cents. This whole decimal thing is pretty annoying tbh.

chmllr (dismissed):

PR #128, opened 2023-05-25
e2e test pipelines

nathanosdev:
This PR superseeds https://github.com/TaggrNetwork/taggr/pull/124.
I opened a new one so I could test the pipeline without spamming notifications on followers of the previous PR.

There's quite a few changes in this PR now, so I'd recommend reviewing by commits.

- https://github.com/TaggrNetwork/taggr/pull/128/commits/6ab8369ad3626ce00955151867a81e2c6c62611d: this commit reverts the revert of the image uploading tests
- https://github.com/TaggrNetwork/taggr/pull/128/commits/3b5dae1c03b6289c54d2ff1114c0ae806a6c7969: this introduces the fix so the docker image fails when the e2e tests fail (this is no longer relevant)
- https://github.com/TaggrNetwork/taggr/pull/128/commits/15f316b9446b003a22634f3b42a032786ebca4f5: this fixes some issues with the e2e tests
- https://github.com/TaggrNetwork/taggr/pull/128/commits/d58162b010b2d0e0c8f55d505a07652a72ee14f3: this moves the e2e test to their own pipeline. I'll leave comments directly on this commit to explain more about what I did there.


webpack.config.js
```diff
@@ -57,6 +57,7 @@ module.exports = {
     output: {
         filename: "index.js",
         path: path.join(__dirname, "dist", frontendDirectory),
+        clean: true,
```
taggrx, line 60: What does it do actually?

taggrx (approved):
Amazing work! Thanks you so much Nathan! Also thanks for comments making all changes so easy to understand!

Btw, did you forget to make `make test` part of `make release` or do you plan to do it in the other PR?

webpack.config.js
nathanosdev: This basically does `rm -rf` on the output directory before building, so I removed this line in favor of this: https://github.com/TaggrNetwork/taggr/pull/128/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L25

nathanosdev:
> did you forget to make make test part of make release or do you plan to do it in the other PR?

I didn't forget, but I think I misunderstood our last conversation about this. I placed the `make test` checks into a separate pipeline here: https://github.com/TaggrNetwork/taggr/pull/128/files#diff-9a979a1e38ba79e2c75e54c4bf21fe1a2a1b935e1736666565f992e634dadd0fR34-R41, with the intention of also adding future checks there for the frontend code too (like eslint, prettier formatting & unit tests etc), pending a conversation about those checks first of course.

Do you want to have these checks as part of `make release` in addition to the additional pipeline? Or instead of the additional pipeline?

taggrx:
Ah I see! No that's fine then. As long as a failing test would break the build, I'm fine with having it as a pipeline job as well, of course!

.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,44 @@
+name: Lint and Test
```
taggrx, line 1: @nathanosdev  Btw why does this needs to install dfx, start it, build a canister, etc?

nathanosdev: The rust compilation fails unless the frontend assets are there and the frontend fails unless the canister IDs file is there. We can skip it by using `NODE_ENV=production` though so it reads the `canister_ids.json` file at the root instead of the local one that's in the `.dfx` folder. Thanks for making me think twice about that! PR here: https://github.com/TaggrNetwork/taggr/pull/142

PR #131, opened 2023-05-27
feat: social image and tags

peterpeterparker:
# Motivation

Improve the rendering of the social cards when the link to taggr is shared on other social platform.

The current card displays the favicon as image  e.g. https://twitter.com/Shivaprasad9989/status/1662350590498971649

# Changes

- fix `raw` url to `apple-touch-icon.png`. note the fix uses `raw` but the IC now supports crawlers, so you should be able to use https://taggr.link/.... However it seems that your React routing is not compliant with static assets therefore still pointed to raw

- add description, `og:` and `twitter:` tags for static html

- provide a social image


src/frontend/assets/social_image.jpg
nathanosdev, line 1: There's a typo in the image, "powered y the" :)

src/frontend/src/index.html
```diff
@@ -3,13 +3,28 @@
     <head>
         <meta charset="UTF-8">
         <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0">
-        <title>TAGGR</title>
         <base href="/">
-        <link rel="apple-touch-icon" sizes="180x180" href="https://6qfxa-ryaaa-aaaai-qbhsq-cai.ic0.app/_/raw/apple-touch-icon.png">
+
+        <title>TAGGR</title>
```
nathanosdev, line 8: This is great, but it should also be updated on the canister side here: https://github.com/TaggrNetwork/taggr/blob/main/src/backend/http.rs#L134-L154

Otherwise now there will be a mix of static description for the `twitter:description` tag and a dynamic description for the `og:description` tag.

nathanosdev (commented):
Welcome to Taggr David! And thank you for your contribution :)

nathanosdev:
@taggrx I approved David for running the CI pipeline. I work with him in Dfinity so I can vouch for him :) he's an excellent frontend developer and works on NNS frontend dapp as well as https://juno.build/.

src/frontend/assets/social_image.jpg
peterpeterparker: Copy/paste used to be more reliable 😫.

Fixed.

src/frontend/src/index.html
peterpeterparker: Does it even work @nathanosdev ? If I inspect the DOM in the browser debugger or with a curl, in both cases I do not get these existing or tag?

If work, should I move any title and description from the static html to the rust backend you pointed out, that's what you mean?

nathanosdev: What curl command are you using?

For me, if I run this then I can see that the dynamic meta tags are added:
```shell
curl https://taggr.link/post/27071
```

I guess you were curling `/`, which does not hit this part of the code because it's already certified. Only the non-certified routes (user posted content) will trigger this as part of the code as part of the "upgrade to update call" process.

So I think it's good to keep what you have added to the `index.html` for the home page, but also add the `title`, `description` and `url` tags to that `http.rs` file so they will be used for the non-certified routes too.

peterpeterparker: Gotcha.

Adding these tags to `index.html` won't break these dynamic non-certified routes right?

In addition to `title`, `description` and `url` we should probably add a `canonical`. Let me try.

peterpeterparker:
```
curl https://taggr.link/post/27071
```

=>

```
<!doctype html><html lang="en"><head><meta charset="UTF-8"><meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1,user-scalable=0"><title>TAGGR</title><base href="/"><link rel="apple-touch-icon" sizes="180x180" href="https://6qfxa-ryaaa-aaaai-qbhsq-cai.ic0.app/_/raw/apple-touch-icon.png"><meta name="apple-mobile-web-app-title" content="TAGGR"><meta name="apple-mobile-web-app-capable" content="yes"><meta content="website" property="og:type"/><meta content="https://6qfxa-ryaaa-aaaai-qbhsq-cai.ic0.app/_/raw/apple-touch-icon.png" property="og:image"/><meta content="https://taggr.link/#/post/27071" property="og:url" />
```

=>

shouldn't the `og:type` be `article` rather than `website` in that case?

peterpeterparker: Pushed few tags, let me know what you think

taggrx:
Thanks so much @peterpeterparker for this contribution! This means a lot to all of us! @nathanosdev yes, I know David from #IC Twitter of course and this came as a big surprise! 😅🙏

@peterpeterparker If you register on Taggr let me know your user name please. We usually reward important contributions with our tokens. They are not transferable yet and can only be obtained in any way except "mining" through content on Taggr or via distribution. Our hope is to achieve a decentralization by putting the governance into the hand of the _right_ people.

taggrx:
social_image
Can we use a compression picture? This one is 75% smaller.

Sorry, I fight for every byte :D

src/frontend/src/index.html
nathanosdev: > Adding these tags to index.html won't break these dynamic non-certified routes right?

You're right, the tags are duplicated for those routes now, the static ones are present in addition to the dynamic ones. I guess we could either try to remove the static ones from the `index.html` in that function for the dynamic metadata OR use a different HTML template to render the dynamic metadata. What do you think @taggrx?

> shouldn't the og:type be article rather than website in that case?

I don't know these meta tags too well so I wouldn't have expected that, but I checked Twitter and see they use the type `article` for Tweets so if it makes sense for them then I definitely think it makes sense for Taggr!

peterpeterparker:
> Can we use a compression picture? This one is 75% smaller.

Sure @taggrx , done (19 KB instead of 80 KB).

peterpeterparker:
Not sure if you wanted to add the source of the image in the repo so here are the source.
I use Affinity Designer and I also exported it as PSD but, this image is nothing else than the existing svg logo, the tagline I found on twitter with the font Helvetica Neue as used by Taggr.

[Archive.zip](https://github.com/TaggrNetwork/taggr/files/11583870/Archive.zip)

src/frontend/src/index.html
nathanosdev: Essentially all of the tags in `http.rs` will be duplicated since they already exist in `index.html` statically.

I was able to replace each individual tag using regex: https://github.com/TaggrNetwork/taggr/commit/8fea4efdaaf65d53296b37c3d65f34ec6367882a

I've tested this for posts, threads, journal, user, realm and feed. It works for all of them with the HTML minified and non minified. But I think this approach will be brittle and we'll need to be careful when changing metatags in the future. It also adds an additional 262kb to the wasm size (the regex implementation for Rust is massive), so the total size is 1,456kb now.

This reminds me of this Stackoverflow post 😆 https://stackoverflow.com/a/1732454

I think that this might be okay as an interim measure, at least until Rust canisters are able to execute JavaScript (I think this is something we'll be able to do later this year). What do you guys think?

taggrx: Ugh, I'd try to avoid the regex dependency at all costs 😅

Why not doing it plain stupid: just substituting of `<meta name="mark" content="OG">` with static tags for the certified call to `index.html` the same way it's done for non certified paths?

nathanosdev: > Ugh

I feel you 😄 

> substituting of <meta name="mark" content="OG"> with static tags for the certified call to index.html the same way it's done for non certified paths?

Ah yes good idea, this can work! I'll give it a try later.

nathanosdev: It works :) 

https://github.com/TaggrNetwork/taggr/commit/c7e1f3139371b2bfdd625a4ab12f262513de34d0

taggrx: @nathanosdev this looks awesome, also a great clean up, thank you so much!

I guess this will be a separate MR?

nathanosdev: I was thinking @peterpeterparker could just cherry pick the commit onto his branch and make any last adjustments, cleanups or additions he thinks to the tags etc.

peterpeterparker:
[social-image.zip](https://github.com/TaggrNetwork/taggr/files/11592451/social-image.zip)

src/frontend/src/index.html
peterpeterparker: I've copied and pushed your code manually in the PR @nathanosdev. Please double check that I did not missed anything.

taggrx:
@peterpeterparker do you want to cherry pick https://github.com/TaggrNetwork/taggr/commit/c7e1f3139371b2bfdd625a4ab12f262513de34d0 so that I can do a final review and approve?

peterpeterparker:
@taggrx I did manually copy the changes see my comment https://github.com/TaggrNetwork/taggr/pull/131#discussion_r1209362572

Or did you add something else?

taggrx:
Oh I am ver sorry, I missed that! Getting too many Github emails :D

taggrx (approved):

taggrx:
Thanks a lot again @peterpeterparker and @nathanosdev !!!

peterpeterparker:
Coolio, thanks for the merge and help

PR #144, opened 2023-05-31
Fixes

taggrx:
Many minor fixes.


src/frontend/src/common.jsx
```diff
@@ -105,7 +105,7 @@ export const getTokens = (prefix, value) => {
     return list;
 };
 
-export const setTitle = value => document.getElementsByTagName("title")[0].innerText = `TAGGR: ${value}`;
+export const setTitle = value => document.getElementsByTagName("title")[0]?.innerText = `TAGGR: ${value}`;
```
nathanosdev, line 108: Ah yes, sorry. This was completely missed by the e2e tests because they run against the replica now. I found this yesterday too and also had a fix on my branch.

nathanosdev (approved):

src/frontend/src/common.jsx
taggrx: No worries, it's a minor issue not affecting the mainnet.

PR #145, opened 2023-06-01
add realms e2e tests

nathanosdev:
This PR adds e2e tests for realms. It also fixes a few bugs in realms. I found more bugs, mostly related to state being out of sync at different parts of the UI but I fixed only the ones that I think are affecting the UX or confusing users in some way. I don't want to spend too much time fixing state syncing issues since I think we can make this much easier by using a state pattern, which would be a much bigger refactoring.

An example of an issue I chose not to fix was after you leave a realm, if you then click the toggle at the top left, the realm that you just left is still present in the list of the users realms. I think it's pretty harmless.

There's also some code that snuck back into the codebase, likely through git branch merges. I'm gonna stop these branch chains now anyway, there's sufficient code here now that I can work on a parallel branch while waiting for a previous branch to be merged.


e2e/pages/global_navigation.ts
nathanosdev, line 1: Another file that somehow got back into the codebase, likely from merging branches again.

e2e/support/post_interaction.ts
```diff
@@ -1,29 +0,0 @@
-import { Locator, Page } from "@playwright/test";
```
nathanosdev, line 1: More code that was deleted in a previous PR and somehow came back.

src/frontend/src/header.jsx
```diff
@@ -73,10 +73,10 @@ export const Header = ({subtle, route, inboxMode, user}) => {
         </div>}
         {loading && <Loading />}
         {inRealm &&
-            <HeadBar title={user.current_realm} shareLink={`realm/${user.current_realm}`} 
+            <HeadBar title={api._user.current_realm} shareLink={`realm/${api._user.current_realm}`}
```
nathanosdev, line 76: This fixes an issue with the `HeadBar` that was happening when a user first enters a realm. None of the data was available so the title did not display and the home link didn't work.

src/frontend/src/realms.jsx
```diff
@@ -139,9 +139,11 @@ export const RealmForm = ({existingName}) => {
                     alert(`Error: ${response.Err}`);
                     return;
                 }
-                await Promise.all([window.reloadCache(), api._reloadUser(), loadRealm()]);
                 if (!editing) {
```
nathanosdev, line 142: There's two issues fixed here. The first is that calling `loadRealm()` here was causing an error popup saying "no realm found" (or something similar) when a new realm was created. I cannot find any impact from removing this, it would be good if you could confirm that I'm not missing some edge case @taggrx.

Second issue was that after creating the realm, the user is redirected to the new realm's page, but the info for the realm is not loaded yet. This makes it look like the user has not joined the realm yet because it shows the "JOIN" button, not the "LEAVE" button.

I saw this happening recently on Taggr but I didn't understand why until I found this bug now. But someone created a realm and did not join it. So I think they created it, were redirected to this page which showed the "JOIN" button, so they clicked it which resulted in them actually leaving the realm since they were already joined, but the UI showed the opposite.

src/frontend/src/header.jsx
taggrx: Amazing!

src/frontend/src/realms.jsx
taggrx: Does the color scheme change still works after without this line? I think it was added to make the oclor changes effective right after saving.

taggrx (approved):

PR #146, opened 2023-06-01
Removes unsafe code

taggrx:
One of the weak spots of Taggr's code base was always the usage of `unsafe` Rust around state variable. This was enough and worked well in the initial MVP phase. But in the current phase, the usage of `unsafe` is irresponsible. It breaks language guarantees and can lead to disastrous bugs that are very hard to find.

For example, this code:

```rust
let s = state();
let p = s.get_reference_to_a_post();
some_async_code_adding_another_post().await;
do_something_with(p, s);
```

can lead to a scary situation, where the hash map holding posts is restructured on the heap and `p` will point to a different post on the heap that the one we retrieved intially, because the post-await part of the code will be only invoked asynchronously and the state could have been changed in between by other messages  arbitrarily.

The refactoring does not add any new features and only introduces an explicit handling of references around asynchronous calls.

I will test this changes in the next days and potentially will add more fixes or improvements.


taggrx:
@nathanosdev is my interpretation correct, that posting with an image is broken on that PR? Somehow, locally I can't reproduce the issue. Locally, images to get uploaded and they work.

taggrx:
Btw, @nathanosdev the e2e tests are 🔥🔥🔥🔥 !!!

src/backend/env/mod.rs
```diff
@@ -1114,12 +1132,13 @@ impl State {
...
-                        proposal.id, err
-                    ));
+                    mutate(|state| {
+                        state.logger.error(format!(
+                            "couldn't vote on NNS proposal {}: {}",
+                            proposal.id, err
+                        ))
+                    });
                 };
-                self.nns_votes += 1;
```
nathanosdev, line 1122: Is this removed intentionally? I see the property is not referenced anymore. Was this intended to be a metric that we would display on the dashboard?

chmllr (approved):
Took me almost an hour today 😅 but LGTM! I like how you can now see different reads and mutations of the state in separate contexts. It's a miracle, we didn't have any bugs yet. Or did we?

nathanosdev (approved):
LGTM 👍 

Glad to see the e2e tests are starting to help catch issues 😄 the hard work is paying off!

And yes, I've also been caught out a few times by misleading exceptions on the tests :(

src/backend/env/mod.rs
taggrx: I didn't remember who added it me or @chmllr and why. And since it's not used anyway, I decided to delete. A number of nns proposals voted on is not saying much to me personally, so I don't think it should be on the dashboard.

PR #152, opened 2023-06-02
Improvements

taggrx:
Minor improvements. Most notable change is a better algorithm for counting popular tags.


chmllr (approved):

src/frontend/src/header.jsx
```diff
@@ -67,28 +68,27 @@ export const Header = ({subtle, route, inboxMode, user}) => {
...
-        {inRealm &&
-            <HeadBar title={api._user.current_realm} shareLink={`realm/${api._user.current_realm}`} 
-                styleArg={{background: realmBg, color: realmFg}}
+        {currentRealm &&
+            <HeadBar title={currentRealm} shareLink={`realm/${currentRealm.toLowerCase()}`} 
+                styleArg={colors}
                 content={<>
-                    <button style={{background: realmBg}} onClick={() =>location.href = `#/realm/${api._user.current_realm}`} >
-                        <Home styleArg={{fill: realmFg}} />
+                    <button style={colors} onClick={() =>location.href = `#/realm/${user.current_realm}`} >
```
nathanosdev, line 80: This is broken again now. If you enter a realm and then click the home icon it will direct you to `/#/realm/null` and show a `404 NOT FOUND` page. I guess it should be `#/realm/${currentRealm}`.

taggrx: Thanks a lot Nathan!

What do you mean with "now"? I was actually fixing the issue that when you enter a realm, the head bar with the color was not taking the assigned color.

nathanosdev: > What do you mean with "now"?

I mean "now" because it was working before the change from `api._user.current_realm` to `user.current_realm`.

PR #154, opened 2023-06-03
wallet e2e tests

nathanosdev:
This PR adds e2e tests for the wallet page.
- ICP transfer in
- ICP transfer out
- Cycle minting


e2e/tests/wallet.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+  icpToE8s,
+  mintingPrincipal,
+} from "../support";
+import { GlobalNavigationElement } from "../elements";
+import { AccountIdentifier } from "@dfinity/nns";
+
+test("wallet", async ({ page }) => {
+  const ledger = await createLedgerClient();
+
+  const user = await test.step("create user", async () => {
```
nathanosdev, line 16: I just discovered this function today. I think it's a nice way to logically break up the long tests into smaller stages. I'll follow up with another PR to refactor the other tests if we agree on this.

package.json
```diff
@@ -16,9 +16,10 @@
   },
   "devDependencies": {
     "@dfinity/identity": "0.15.5",
-    "@dfinity/nns": "0.15.1",
+    "@dfinity/nns": "0.16.0",
+    "@dfinity/cmc": "0.0.11",
     "@dfinity/principal": "0.15.5",
-    "@playwright/test": "^1.32.3",
+    "@playwright/test": "1.34.3",
```
nathanosdev, line 22: After updating playwright, `npm run install:e2e` needs to be run again locally to get the updated system dependencies (pretty just the browser binaries).

taggrx (approved):

PR #156, opened 2023-06-03
refactor e2e tests to use steps

nathanosdev:
This PR refactors e2e tests to use Playwright's step functionality. This should make the test organization easier to reason about and better isolate the individual stages of long-running tests. I've also reduced some duplication in the tests.


taggrx (approved):
NICE!

PR #157, opened 2023-06-04
small metadata optimizations

nathanosdev:


src/frontend/assets/hash.svg
taggrx, line 1: What has changed here actually?

src/frontend/assets/logo.min.svg
taggrx, line 1: And here

src/backend/assets.rs
```diff
@@ -70,6 +70,15 @@ pub fn load() {
         include_bytes!("../../dist/frontend/apple-touch-icon.png").to_vec(),
     );
 
+    add_asset(
```
taggrx, line 73: I don't understand this part. Why is it needed? Is it loaded somewhere at all?

webpack.config.js
```diff
@@ -76,6 +75,16 @@ module.exports = {
         new HtmlWebpackPlugin({
             template: path.join(__dirname, asset_entry),
             cache: false,
+            minify: isDevelopment ? false : {
```
taggrx, line 78: Nice! Any noticeable improvements in frontend code size?

src/backend/assets.rs
nathanosdev: It's set as the image for socials in the `index.html`: https://github.com/TaggrNetwork/taggr/blob/main/src/frontend/src/index.html#L8-L10

So when socials try to load that image it doesn't work and it looks like this:
image

src/frontend/assets/hash.svg
nathanosdev: I had tried to optimize these with https://jakearchibald.github.io/svgomg/ but it caused issues so I decided to defer to a later date and then when I reverted the changes I deleted newlines from the original file by accident. I'll revert properly now.

webpack.config.js
nathanosdev: Nothing worth writing home about 😆 

2.54kb -> 2.25kb

chmllr (approved):

taggrx (approved):

PR #158, opened 2023-06-04
add e2e test for journal

nathanosdev:


src/frontend/src/form.jsx
```diff
@@ -64,6 +64,7 @@ export const Form = ({postId = null, comment, realmArg = "", expanded, submitCal
             };
             await submitCallback(value, blobArrays, extension, realm);
             setValue("");
+            clearTimeout(choresTimer);
```
taggrx, line 67: Oh! Interesting! was it a leak actually? Or what happens to a timer when the view is destroyed?

taggrx (approved):

src/frontend/src/form.jsx
nathanosdev: The timer is independent of the view since it's already sitting somewhere on the JavaScript event loop, so what happens here is the value of the input was set to the local storage after the post was already saved. So when the tests were creating multiple posts in a row, _sometimes_ the previous post content would be still in the `textarea` for creating a new post. It can only happen if you're really, really quick though (like the e2e tests) 😄 I don't think any human user could possibly run into this issue.

PR #165, opened 2023-06-05
make upgrades safe

chmllr:
Last weekend we had a critical error on one of the subnets related to call context accounting and while I was reading the analysis done by other engineers, I realized that the current upgrade mechanism used in the Taggr canister is not safe. Consider the scenario where the main canister sends the new binary to the management canister, but before the management canister upgrades the main canister, an ingress message triggers a canister call. At that moment there is no one to check if the main canister has a new open call contexts and the upgrade will be executed. However the callback of the open call might then corrupt the heap once the response arrives.

I propose that before any upgrade, we store the current time and panic on any ingress trying to initiate a canister call within 5 minutes after the upgrade attempt. It could be also just a simple boolean flag instead, but then we would need an external way to unset the flag if something goes wrong in an unforeseen situation and the upgrade does not happen (and the timestamp won't get reset). With a timestamp, the canister will simply unlock itself in 5 minutes. Any upgrade should happen in a matter of seconds.

I also renamed the accounting functions to make them sound like declarations and not predicates, but I can change the names back if somebody feels strongly about it.


taggrx:
This is a great catch! Thanks so much Christian! I actually was thinking about this once, that I do not understand how atomic the upgrade execution is, but decided to investigate later...

taggrx (approved):
Tested upgrades locally, it works.

src/backend/env/canisters.rs
```diff
@@ -19,22 +19,32 @@ use serde::{Deserialize, Serialize};
...
 
 thread_local! {
     static CALLS: RefCell<HashMap<String, i32>> = Default::default();
+    // A timestamp of the last upgrading attempt
+    static UPGRADE_TIMESTAMP: RefCell<u64> = Default::default();
 }
 
-pub fn call_opened(id: &str) {
+pub fn open_call(id: &str) {
+    // Before opeining a new casniter call, we first check whether an upgrade was intiated within the last 5 minutes and panic if it is the case.
```
nathanosdev, line 33:
```suggestion
    // Before opening a new canister call, we first check whether an upgrade was initiated within the last 5 minutes and panic if it is the case.
```

nathanosdev: `cansiter` is the bane of my life. I can't count how many times I've made that mistake.

nathanosdev (approved):

nathanosdev:
Could we use a post upgrade hook to reset a boolean variable instead of relying on 5 minutes being enough? I don't know how long upgrades usually take or if it's even possible for it to take more than 5 minutes, but are we also potentially creating too large of a window where we certain canister functionality will fail?

Granted I think it won't be an issue now, but it could become an issue as the user base (hopefully) grows.

src/backend/env/canisters.rs
chmllr: Same here! I thought it's me like something wrong with my reflexes when typing this word... 😂

nathanosdev:
> I think you misunderstand.

Yes I misunderstood, thanks for explaining. Now it makes sense!

PR #166, opened 2023-06-05
Post archiving

taggrx:
This is the final PR closing my series of experiments and introducing the ultimate scalability for posts. After this is deployed, the Taggr canister will only hold 20k last created or modified posts in the heap. Everything else will be offloaded into the stable storage every day. This allows us to have a bounded heap size at any moment and avoid running into instruction limits on upgrades.

Now, we do not need to worry about posts until their collective size is under 48 Gb.


src/backend/env/post.rs
```diff
@@ -4,6 +4,26 @@ use crate::mutate;
...
+static mut CACHE: Option<BTreeMap<PostId, Box<Post>>> = None;
+
+// This is a static cache that is only populated during one queries's life cycle. The reason why it's
+// needed is that `Post::get` always returns a reference to a post. But when we load an archived
+// post, we read raw bytes from the stable memory and deserialize them into a value. To return a
+// reference to that value, we need to anchor it somewhere on the heap. This is where we need our
+// cache. But we also _cannot_ store posts inside the cache directly, because any mutation of the
+// cache within a query life cycle will restructure the hash map and hence break all references.
+// To work around this issue, we go through one level of indirection and box all posts (put them on
+// to the heap) and then add only add the pointers to the boxed value into the cache. This way,
```
chmllr, line 16:
```suggestion
// to the heap) and then only add the pointers to the boxed value into the cache. This way,
```

src/backend/env/post.rs
```diff
@@ -527,30 +554,81 @@ impl Post {
     }
 
     pub fn count(state: &State) -> usize {
-        state.posts.len()
+        state.posts.len() + state.memory.posts.len()
     }
 
+    // Get the post from the heap if available, or load from the stable memory into the cache and
+    // return the refernce to it
```
chmllr, line 561:
```suggestion
    // return the reference to it
```

src/backend/env/post.rs
```diff
@@ -527,30 +554,81 @@ impl Post {
...
 }
 
+// Moves a configured number of posts from hot to cold memory.
+pub fn archive_cold_posts(state: &mut State, max_posts_in_heap: usize) -> Result<(), String> {
+    let mut posts: Vec<(u64, PostId)> = state
+        .posts
+        .values()
+        .map(|post| (post.timestamp(), post.id))
+        .collect();
+    posts.sort_unstable_by(|(t1, _), (t2, _)| t1.cmp(t2));
```
chmllr, line 611: I'd just get the timestamp in the sorting function
```suggestion
        .collect();
    posts.sort_unstable_by(|(p1, _), (p2, _)| p1.timestamp().cmp(p2.timestamp()));
```

taggrx: Good idea, thanks Christian!

nathanosdev (approved):

src/backend/env/post.rs
```diff
@@ -6,6 +6,26 @@ use crate::mutate;
 use crate::reports::Report;
 use serde::{Deserialize, Serialize};
 
+static mut CACHE: Option<BTreeMap<PostId, Box<Post>>> = None;
```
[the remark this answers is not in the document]
taggrx: Unfortunately, functions are not allowed in static variables initializations, it can only be a value. So an Option here is a workaround, unfortunately.

PR #170, opened 2023-06-07
ICP tipping

taggrx:


docs/WHITEPAPER.md
```diff
@@ -104,11 +104,6 @@ Note:
 2. The karma and cycles of every user inactive for longer than `$inactivity_duration_weeks` weeks decreases by `$inactivity_penalty` per week.
 3. Users with a negative karma do not participate in distributions.
 
-### Tipping
```
taggrx, line 107: Removed becasue I think we don't need to describe ICP tipping. Before it was different, it affected karma first, then it was only cycles which is a weird tipping anyway, but now it's obvious.

chmllr (approved):
So no refund on post deletion anymore? 😂

nathanosdev (dismissed):
Awesome!

PR #171, opened 2023-06-08
Fixes

taggrx:
Contains a couple of self explanatory fixes and improvements. The largest one is dropping of the post index on each users profile. It was a premature optimization.


chmllr (approved):

src/backend/env/mod.rs
```diff
@@ -1510,15 +1511,15 @@ impl State {
         Box::new(
             posts
                 .filter_map(move |i| Post::get(self, &i))
-                .filter(move |post| with_comments || post.parent.is_none()),
+                .filter(move |post| !post.is_deleted() && (with_comments || post.parent.is_none())),
```
nathanosdev, line 1514: When a post is deleted, is the karma returned to the original giver?

taggrx: All rewards are refunded as cycles, if that's what you're asking. The karma of the post author is also decreased by at least the amount of rewards they received.

nathanosdev: Yes that's perfect, thanks!

PR #175, opened 2023-06-10
add release proposal e2e tests

nathanosdev:


docs/LOCAL_DEVELOPMENT.md
```diff
@@ -144,6 +144,26 @@ To update screenshots/snapshots:
...
+```shell
+dfx canister call taggr demigodmode '("${username}")'
+```
+
+Return a user to bootcamp:
+
+```shell
+dfx canister call taggr peasantmode '("${username}")'
+```
+
```
nathanosdev, line 166: We need to clean up the trusted users after tests, otherwise repeated runs against the local replica will fail because the token distribution will shift. It's 1% of VP required to accept a proposal, so it's 99% to reject, so we need to vote with all members in the test to reject a proposal. This problem would still exist even if we change the required VP though.

src/backend/lib.rs
```diff
@@ -116,18 +116,60 @@ mod dev {
 
     #[update]
     // Promotes any user to a stalwart with 20k tokens.
-    async fn godmode(user_id: UserId) {
+    async fn godmode(username: String) {
```
nathanosdev, line 119: I opted for username here because that's easier to get for the e2e tests

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -144,6 +144,26 @@ To update screenshots/snapshots:
...
+Make a user trusted:
+
+```shell
+dfx canister call taggr demigodmode '("${username}")'
+```
+
+Return a user to bootcamp:
+
+```shell
+dfx canister call taggr peasantmode '("${username}")'
```
taggrx, line 164: Great names! 🤣

e2e/tests/proposal.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+        const proposalsPage = await globalNavigation.goToProposalsPage();
+        const commitHash = generateGitCommitHash();
+        const binaryPath = resolve(
+          __dirname,
+          "..",
+          "..",
+          "target",
+          "wasm32-unknown-unknown",
+          "release",
+          "taggr.wasm.gz"
```
taggrx, line 74: If I understand correctly, this is only testing that the proposal goes through, not that the upgrade was executed, right? Can we make one more small step to actually test that the upgrade was a success? For that we'd need to trigger `dfx canister call taggr chores` after the voting - this is a dev method that triggers the hourly routine and executes approved upgrades. The only challenge would be to have a binary with a different hash so that we can compare with `dfx cansiter info taggr` and make sure the canister is running a new hash. What do you think about it? How hard would it be?

taggrx (approved):
Wow thank you so much Nathan!!! 🤯🔥

I left one small comment, please let me know what you think?

e2e/tests/proposal.spec.ts
nathanosdev: Yeah it's not actually testing the upgrade and I agree we should definitely test that.

Getting a different binary is the biggest challenge there I think. Everything else should be trivial.
Running this test in parallel with other tests might break them, but we could work around that by creating a different "suite" of tests for testing the upgrade.

taggrx: LEt me then create a separate bash script based test for that later.

src/backend/lib.rs
```diff
@@ -116,18 +116,60 @@ mod dev {
...
         mutate(|state| {
-            let user = state.users.get_mut(&user_id).expect("no user found");
-            user.timestamp -= CONFIG.trusted_user_min_age_weeks * WEEK;
+            let user = state
+                .users
+                .values_mut()
+                .find(|user| user.name == username)
+                .expect("no user found");
```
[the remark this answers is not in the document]
nathanosdev: I made an attempt at this, please let me know if I'm abusing something 😆

src/backend/lib.rs
```diff
@@ -120,21 +120,57 @@ mod dev {
...
             let principal = user.principal.clone();
             token::mint(state, account(principal), CONFIG.max_funding_amount);
-        })
+        });
+    }
+
+    #[update]
+    // Promotes any user to trusted status with 20k tokens.
+    async fn demigodmode(username: String) {
+        let user_id = read(|state| state.user(&username).expect("no user found").id);
```
taggrx, line 142: You can pull it into the `mutate` closure if you dereference it:

    let user_id = *state.user(&username).expect("no user found").id;

Or just don't worry, I can do it myself later, it's a tiny nit in case of these methods.

taggrx (approved):

taggrx (approved):

PR #176, opened 2023-06-12
Fix colors on android

taggrx:
I had a chance to check out Taggr on Android this weekend. I think this should improve the colors for the app.


chmllr (approved):

nathanosdev (approved):

taggrx:
I added a tiny fix on top: moving out of realms should only work on root posts.

taggrx:
It does, but I have admin priveleges which I can't disable and sometimes misuse 😂 for cases where it's obviously not breaking anything and even if it is, we'll see a broken master and no release can be created. This is the only reason I allow myself to do it to save time.

taggrx:
And of course I can stop doing that if it makes you uncomfortable, really, no problems with that.

taggrx:
Even worse 🙈 I sometimes push directly to master, like I just did: https://github.com/TaggrNetwork/taggr/commit/a674f91c088caa831745fd3a105f2218c4b97504 

But this is more to not annoy you guys with my stupid mistakes. And because when comparing the releases all changes are visible anyway.  I always feel bad for pushing so many changes through you 😏

nathanosdev:
You really shouldn't feel bad for pushing changes through. By spreading the responsibility for the deployed changes among a few developers we're making the Taggr platform, DAO, and community safer.

I was just curious how it was going through anyway. As long as the build passes on master before a release is proposed then I'm happy and my main reason for that is to just make sure that the DAO doesn't suffer our silly mistakes :)

PR #179, opened 2023-06-12
begin typescript migration

nathanosdev:


webpack.config.js
```diff
@@ -1,7 +1,6 @@
 const path = require("path");
 const webpack = require("webpack");
 const HtmlWebpackPlugin = require("html-webpack-plugin");
-const TerserPlugin = require("terser-webpack-plugin");
```
nathanosdev, line 4: I was supposed to remove this awhile back, must have crept back in with git merges 🙄

src/frontend/src/api.ts
```diff
@@ -0,0 +1,121 @@
...
+    const query_raw = async (canisterId = defaultCanisterId, methodName: string, arg = new ArrayBuffer(0)): Promise<ArrayBuffer | null> => {
+        let response = await agent.query(canisterId, { methodName, arg }, identity);
+        if (response.status != "replied") {
+            console.error(response.status);
+            return null;
+        }
+        return response.reply.arg;
+    };
+
+    const query = async <T>(methodName: string, arg0: unknown, arg1: unknown, arg2: unknown, arg3: unknown, arg4: unknown): Promise<T | null> => {
```
nathanosdev, line 25: I think we can do better than `unknown` here, but I didn't want to change this too much on the first PR.

@taggrx I noticed there's a mix of JSON encoding and Candid encoding here. What's your preferred approach?

Candid has some nice static typing properties and code generation tools that we could leverage and then remove about 90% of this file if we went that route.

Of course, JSON is the more natural conclusion for JavaScript and there are also static typing/code generation tools we can use there (like Swagger/OpenAPI) but the IC needs a bit more tooling in place before we can use stuff like that. I'd love to be able to use an approach similar to actix-web eventually, but that will take some time.

webpack.config.js
taggrx: Happens to the best of us!  😅

src/frontend/src/api.ts
taggrx: Candid turned out to be utterly useless for me except for those cases where I need to pass over a binary. Those are the only cases where I am still using it, because AFAIK there is no way for efficient binary data encoding using JSON.

taggrx (commented):
Thanks so much Nathan! 🤯🔥 This looks amazing! Btw, did you actually do this?

    git mv src/frontend/src/api.js src/frontend/src/api.ts

?

If not I wonder if the diff would be more readable?

src/frontend/src/api.ts
nathanosdev: Well yes, if it's only used as an encoding then it's no different than JSON, aside from the binary encoding.

But is static type generation utterly useless?
It's still a WIP but this PR https://github.com/dfinity/cdk-rs/pull/386 will allow the `.did` file to be autogenerated from the Rust code and then we could autogenerate TypeScript code from the `.did` file. Then we can import and use type safe actors to interact with the canister without having to hand write all of the interfaces in TypeScript.

nathanosdev:
> git mv src/frontend/src/api.js src/frontend/src/api.ts

I didn't do this originally but I tried it now and unfortunately it didn't help :(

> I am curious whether the timeout increases are related to TS?

They're not, I just bundled them in but I've moved them out to improve confidence in the TS changes.

src/frontend/src/api.ts
taggrx: So 1.5 years ago the tooling looked like this: it would generate some JS code which sometimes wasn't updated at all and I have spent hours debugging problems only to find out that for whatever reason the auto-generation part of the candid interface didn't work again. Then, not being able to debug anything in the JS console was very painful as well. Not using TypeScript didn't provide any type safety for me in the first place. But now, if this tooling has matured and after we switch to TS it might be a different story indeed. So I am obviously not against type safety, the opposite!

nathanosdev:
So it seems that the issue actually is related to the Typescript changes...
I'm not sure how that's possible with the changes I made, so I think we shouldn't merge this until I figure out why that's happening.

src/frontend/src/api.ts
nathanosdev: I see, yes that indeed does sound like an absolute nightmare. I can give it a try at some point later and see if the experience is better. So far on my personal projects it works well, but maybe there's something I'm not considering on Taggr.

nathanosdev:
😩 nevermind. I was able to reproduce the failures on `main` before applying the branch with the increased timeouts. I also stepped backward through the TypeScript changes to identify anything causing issues with the e2e tests but they just failed or passed randomly, so I don't believe it was anything to do with the TypeScript changes.

taggrx (approved):

taggrx (approved):

PR #188, opened 2023-06-13
add prettier support

nathanosdev:
Introduces support for Prettier, with a pre-commit hook and a lint check on the pipeline.
I apologize in advance for this, the changes are absolutely diabolical.
Much worse than I expected.

It's probably not realistic to actually review these changes line by line. But Prettier is a really mature tool at this point, so I think the odds of it breaking something are really low. The e2e tests passing give some confidence too.

I haven't found any issues testing manually either but it would be awesome if you could give it a once-over as well @taggrx.


.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
```
nathanosdev, line 1: For reference: https://editorconfig.org/

Prettier can be configured using a `.prettierrc` file, but editors won't necessarily reflect the same settings as Prettier. So using this file, we can install the EditorConfig plugin so that the IDE matches this config and Prettier will also automatically read this config too.

I'm also not too opinionated about the settings so I'm open to suggestions. I generally prefer 2 spaces, but `cargo fmt` forces 4 spaces for Rust so I think we might as well keep that consistent.

I went for 2 spaces for yaml because even though it's not supposed to make a difference, I've had issues in the passed using 4 spaces (or tabs) for yaml.

.gitignore
```diff
@@ -1,5 +1,6 @@
 # Various IDEs and Editors
-.vscode/
+.vscode/*
+!.vscode/extensions.json
```
nathanosdev, line 3: I've allowed this file to be committed so anyone using VSCode can get a set of recommended plugins out of the box that can improve their developer experience.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,5 @@
+repos:
```
nathanosdev, line 1: Using pre-commit is optional, but it can save a lot of time. There's nothing more annoying than waiting ages for a pipeline to finish only to find out that it failed because of the formatting.

src/backend/metadata.rs
```diff
@@ -49,7 +49,7 @@ pub fn set_metadata(
     .replace('\n', "");
 
     String::from_utf8_lossy(body)
-        .replace(r#"<meta name="mark" content="OG">"#, &metadata)
+        .replace(r#"<meta name="mark" content="OG" />"#, &metadata)
```
nathanosdev, line 52: This is one of two actual code changes I needed to make. Prettier adding the trailing slash in the `index.html` file.

src/frontend/src/common.jsx
```diff
@@ -1,136 +1,238 @@
...
         })
-        .map(token => token.match(regexp)[0]);
+        .map((token) => token.match(regexp)[0]);
     const list = [...new Set(tokens)];
     list.sort((b, a) => a.length - b.length);
     return list;
 };
 
-export const setTitle = value => document.getElementsByTagName("title")[0]?.innerText = `TAGGR: ${value}`;
+export const setTitle = (value) => {
```
nathanosdev, line 181: This is the other actual code change I had to make. Prettier didn't like the syntax that was originally used.

PR #199, opened 2023-07-01
setup separate canister for e2e tests

nathanosdev:
This PR setups a separate canister for running e2e tests. The canister will always start in a fresh state. This allows the e2e tests to have a predictable starting state without interfering with a long-running canister that a developer might use for local development. This will also make the DAO focussed tests easier because we won't need to worry about each test creating new DAO members and shifting the VP. I haven't changed anything there yet, but this will allow that code to be simplified. I'll follow up in a future PR to address that.

I've also added some minor tweaks to improve the stability of the e2e tests.

Additionally, I'm opening this PR prematurely. It's dependent on another PR being merged, the package being published, and the reference in the `package.json` being updated on this branch.https://github.com/dfinity/ic-js/pull/373, the build will fail until then. I'll be away for 2 weeks now so I won't follow up on this until I'm back. @taggrx if you want to merge this in the meantime, could you keep an eye on the other PR and update this one once the package is published? @peterpeterparker will be the one to publish that. If not, I can handle it once I'm back.


src/frontend/src/new.jsx
```diff
@@ -48,6 +48,7 @@ export const PostSubmissionForm = ({ id, repost }) => {
             }
             postId = result.Ok;
         }
+        await window.reloadCache();
```
nathanosdev, line 51: When we add or edit a post, there's a possibility of creating a new storage canister. This issue was exaggerated by the e2e tests always starting from a fresh state which caused the image uploading test to always fail. In reality, this will be an extremely rare event.

src/frontend/src/post.jsx
```diff
@@ -364,7 +364,7 @@ export const Post = ({
                         classNameArg="post_extension repost"
                     />
                 )}
-                {showExtension && post.extension.Proposal && (
+                {showExtension && "Proposal" in post.extension && (
```
nathanosdev, line 367: Similarly, this was a new issue caused by running the tests always in a fresh state. The first proposal's ID is 0 (hence falsey) so the previous check that we had here would fail and the test would always fail. I hate JavaScript...

peterpeterparker:
Just added few minor comments to your PR @nathanosdev.
Sure thing, I can release a new versionof ic-js begin of next week 👌

nathanosdev:
Ah awesome, thanks @peterpeterparker! I didn't expect comments so soon 😅 I've addressed your suggestions now.

src/frontend/src/post.jsx
taggrx: I hate JS... 😂 Thanks Nathan!

peterpeterparker:
@nathanosdev lib has been released (`v0.0.4`)

src/frontend/src/new.jsx
taggrx: I debugged this issue. It's 100% reproducible on a new post without storage buckets. The query fetching the blob on the created post is failing with this error:

```js
Object { status: "rejected", error_code: "IC0301", reject_code: 3, reject_message: "IC0301: Canister tzq7c-xqaaa-aaaaa-aaamq-cai not found" }
​
error_code: "IC0301"
​
reject_code: 3
​
reject_message: "IC0301: Canister tzq7c-xqaaa-aaaaa-aaamq-cai not found"
​
status: "rejected"
​
<prototype>: Object { … }
```

But this is clearly a dfx issue. Since Taggr _knows_ already the canister id (the post creation blocks on canister creation) the canister definitely exists by the time we try to fetch the blob from it! But the internal router of `dfx` is not updated fast enough.

So reloading of the cache _might_ help by introducing a delay. I don't think it's needed per se in the production code, the delay could be introduced in the test, right?

taggrx: I was wrong, it's not a dfx issue. After we create the very first post a bucket canister gets created, yet its id is unknown on the frontend yet and this is why cache reloading fixes it (by loading it).

I'll add a fix in my current PR. It's basically the same, just conditioned on empty bucket list and non-empty blobs. We don't want to always wait, right.

nathanosdev: > After we create the very first post a bucket canister gets created, yet its id is unknown on the frontend yet and this is why cache reloading fixes it

Yes, sorry I should have explained that better in my original comment.

> It's basically the same, just conditioned on empty bucket list and non-empty blobs

Yeah, I think this approach is better. Someone could still run into this on mainnet if their post causes a new bucket to be created, but that's going to be extremely rare.

src/backend/lib.rs
```diff
@@ -166,6 +166,26 @@ mod dev {
         });
     }
 
+    #[update]
+    // sets the controller of all storage buckets.
+    // allows the caller to manage storage buckets for cleanup.
+    async fn set_buckets_controller(principal: String) {
```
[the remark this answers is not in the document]
nathanosdev: Ah that's much cleaner, thanks!

e2e/tests/post_creation.spec.ts
```diff
@@ -171,6 +171,8 @@ test("post creation with image", async ({ page }) => {
         });
 
     await test.step("check uploaded image", async () => {
+        // wait for image to load, find a more reliable way to handle this
```
taggrx, line 174: Do we still need this? I thought this is now solved by https://github.com/TaggrNetwork/taggr/blob/main/src/frontend/src/new.jsx#L52

So in the post creation handler, if we detect at least one blob but no buckets yet, then this is the first ever message creating a bucket.

I think I'll move this logic later into the canister initialization making sure we always have at least one bucket...

taggrx (approved):
Thanks so much for this improvements and welcome back! 😎🔥

e2e/tests/post_creation.spec.ts
nathanosdev: This is a different issue that's present in the same part of the flow. The image uploading e2e test is failing randomly because Playwright is not waiting for the image to load before running the assertions. I tried a few different things already to make it wait but haven't found anything reliable.

I even tried this nonsense: https://github.com/microsoft/playwright/issues/6046#issuecomment-1532438713... Maybe if we switched to loading image by an `https:` URL instead of `blob:` then this hack would work, but I'd rather try find some cleaner method. I'll continue to revisit the flaky tests while I work on other things to see if I can find better ways to improve their stability.

nathanosdev:
@taggrx can you merge when you get a chance? :)

e2e/tests/post_creation.spec.ts
taggrx: Sorry, I responded to a wrong line apparently. I meant this one: https://github.com/TaggrNetwork/taggr/pull/199/files#r1248909623

PR #201, opened 2023-07-03
Subsidized invites

taggrx:
Introduces subsidized invites (up to 300 cycles weekly).


src/backend/env/mod.rs
```diff
@@ -554,21 +554,32 @@ impl State {
...
+                            new_user_id,
+                            cycles,
+                            0,
+                            Destination::Cycles,
+                            "claimed by invited user",
+                        )
+                        .map_err(|err| format!("couldn't use the invite: {}", err))?;
+                    new_user_id
+                } else {
+                    return Err("inviter has not enough funds".into());
```
chmllr, line 578:
```suggestion
                    return Err("inviter has not enough cycles".into());
```

chmllr (approved):

src/backend/env/mod.rs
taggrx: Thanks a lot!

PR #203, opened 2023-07-07
introduces total user ICP account

taggrx:
I recently realized that the current way of keeping users' funds in the Treasury until withdrawal would lead to problems when the Treasury runs low because the rest of the code, when deciding if a distribution should take place or not, only looks at the total amount of ICP in the Treasury.

In this change we introduce a second subaccount of Taggr which will contain all user ICP to be withdrawn, separately from the Treasury. This also fixes the treasury number on the dashboard.


chmllr (approved):
LGTM

PR #207, opened 2023-07-22
add design system boilerplate

nathanosdev:
This is a bit experimental and not immediately useful but I wanted to get some thoughts/feedback on this before pursuing it further. This is essentially the barebones of a Taggr design system and there are a few reasons why I think this could be useful for Taggr.

Cleaner code - In a general sense, I find the thought process of writing code for multiple consumers produces much cleaner code. UI component libraries have been a particularly beneficial example of this for me in my experience. The responsibility of page layout is further separated from individual component design. Refactoring data flow into a state management library would be another example of this that worked well for me in the past and I plan to start looking into that now too.

Refactoring - The Taggr frontend needs some more design patterns in my opinion, but refactoring JavaScript is particularly dangerous. Pulling out reusable components is a relatively simple form of refactoring that gives us the opportunity to more safely convert them to TypeScript and potentially rethink the design of the components with a clean slate. This also reduces the potential bug surface for future refactors since the main frontend codebase would be smaller.

Reuse - In the context of creating alternative frontends, having a library of framework-agnostic and flexibly themeable components will make experimentation with different forms of the Taggr interface much faster. Particularly when it comes to more complex components like the post editor.

I've opted to try out [lit](https://lit.dev/) for making these components since it's an extremely small library and it leverages browser native [web components](https://www.webcomponents.org/). This means the components will be framework agnostic and can be used with React, Angular, or any other frontend framework. The same effect could be used by leveraging the web components API directly, but it's nasty. Lit helps to make it much nicer to work with.

I've also taken a different approach to styling the components. [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/var) are used extensively to make the components as flexible as possible. Functionality, interactions, and basic styling can be reused while allowing for "coats of paint" to be easily added on top. I also moved the calculation of some theme variables into [SASS](https://sass-lang.com/). This allows a static theme stylesheet to be produced at compile time that can be swapped at runtime.

One last thing I've added is [Storybook](https://storybook.js.org/). It's an environment for developing components in isolation, without having to run an application. It can also serve as a testing ground and documentation for components. There are tons of useful addons, like for measuring accessibility conformance or adding controls for replicating component state while testing and some other cool things. I've already added a theme switcher which I think will be really handy for seeing components in different themes really quickly.

To try some of this out then an `npm i` will be required to install the new packages, then `npm run build -w @taggr/ui` will build the UI component library, and `npm run storybook -w @taggr/ui` will run the Storybook server (it will auto open a browser).


taggrx:
Will respond as soon as I get the chance to carefully read and review the PR but it sounds absolutely great initially!

taggrx:
After looking closer at the proposed changes I do feel like it's a great improvement! For me, as a non-frontend engineer some of these things a bit difficult to get intuitively because I just don't have enough experience in writing frontends. 

For example, one thing I still didn't really get is the lit library. Why does a framework agnostic abstraction adds value? In theory, I guess we become less dependent on React, is that the case? The example with the button didn't somehow make it clear to me what we had before and what we get with the lit library. I'd appreciate if you could help me understand that part.

nathanosdev:
And also btw, if the framework agnosticism does not feel like a worthwhile trade off with the added tech and complexity then I'm fine with writing these components in React.

taggrx:
> This is more valuable for anyone using the components to build an alternative frontend for Taggr than it is for us with the current frontend

Ok, this is a really good argument!

So if you think you'd translate all components eventually to lit, I think we should do it. (I think I couldn't help initially and would need to see a couple of examples.)

nathanosdev:
Great! I'll clean the PR up and get the build passing over the weekend.

nathanosdev:
Sorry I left this open for so long. I'll close this PR, keep the branch on my fork and follow up once I have capacity to actually follow through on it :)

PR #212, opened 2023-07-31
Automated top up from revenue and rewards up to 1000 cycles

taggrx:


src/backend/env/mod.rs
```diff
@@ -970,7 +970,9 @@ impl State {
             .values_mut()
             .filter(|u| u.karma_to_reward() > 0)
             .filter_map(|user| {
+                let _ = user.top_up_cycles_from_rewards();
```
chmllr, line 973: So it'll happen once per week and up to 1k cycles, right?

chmllr (approved):

src/backend/env/mod.rs
taggrx: Yep!

PR #216, opened 2023-08-07
Next

taggrx:
- Fixes throwing aways of too many NNS proposals (rejects them instead)
- Improves the release profile for canisters and upgrades the bucket.


chmllr (approved):
LGTM!

PR #219, opened 2023-08-12
fix: typos

omahs:
fix: typos


taggrx:
Thanks so much @omahs! What's your name on Taggr?

PR #226, opened 2024-01-05
Fix share profile link

leo-ship-it:
The link you get to your clipboard when copying your profile is broken.


taggrx:
Thanks so much Leo! 🙏 However, the proper fix should have no back-slash at all, then it's a proper sharable link with SEO data in the html file. You can check yourself the output of `curl https://taggr.link/user/leo`. I'll include this in my current set of changes.

leo-ship-it:
Ok, that's what I thought at first but I saw that you were using `/#/wallet` for example so I guessed you somehow wanted to have this `/#/` pattern for some reason.

PR #228, opened 2024-01-11
chore: review comments

taggrx:
Addresses comments on the last release: unit test + minor refactoring.


src/backend/updates.rs
```diff
@@ -114,34 +114,21 @@ fn vote_on_report() {
...
 fn link_cold_wallet(user_id: UserId) -> Result<(), String> {
-    mutate(|state| {
-        let user = state.users.get_mut(&user_id).ok_or("no user found")?;
-        if user.cold_wallet.is_some() {
-            return Err("this user has already a cold wallet".into());
-        }
-        user.cold_wallet = Some(caller());
-        Ok(())
-    })
+    mutate(|state| state.link_cold_wallet(caller(), user_id))
```
ulan, line 126: Refactoring looks good. A question about the flow here: shouldn't the user also approve this change?

ulan (approved):
lgtm

src/backend/updates.rs
taggrx: Good question, I also thought about it but decided that since this adds and not subtracts, it's the same as asking the user approving the reception of funds, isn't it? Means: receiving funds does not prove anything, yet you can use it.

Does it make sense?

PR #229, opened 2024-01-12
Security improvements

taggrx:
- Fixes a potential rewards lost during withdrawals.
- Skips users for giving rewards if their balance is negative.
- Adds minting safe guards.


src/backend/env/mod.rs
```diff
@@ -1050,6 +1046,8 @@ impl State {
             }
         }
 
+        self.minting_mode = false;
```
ulan, line 1049: Nit: feel free to ignore.

This is a bit fragile because if an early return is added in the code above, then clearing of the flag would not run.
Suggestions: 
- extract the code between setting and clearing the flag into a separate function.
- consider https://docs.rs/scopeguard/latest/scopeguard/

src/backend/env/mod.rs
```diff
@@ -1628,24 +1626,33 @@ impl State {
...
+            state.principal_to_user_mut(principal).map(|user| {
+                let cloned_user = user.clone();
+                user.treasury_e8s = 0;
+                cloned_user
+            })
+        })
+        .ok_or("no user found")?;
+
         let fee = invoices::fee();
+        let rewards = user.treasury_e8s;
```
ulan, line 1639: Nit: looks like we only need `treasury_e8s` and `principal` from the user. Maybe instead of cloning the user, we could return just that pair from the helper above?

src/backend/env/mod.rs
```diff
@@ -1628,24 +1626,33 @@ impl State {
...
-            });
+            .await
+            {
+                mutate(|state| {
+                    if let Some(user) = state.users.get_mut(&user.id) {
+                        user.treasury_e8s = rewards
+                    }
+                });
+                return Err(err);
+            }
```
ulan, line 1655: Note that in the case when `if Tokens::from_e8s(user.treasury_e8s) > fee` doesn't hold, we are not going to restore user's `treasury_e8s`. It is intended? If so, maybe add a comment about it in the `else` branch?

src/backend/env/mod.rs
```diff
@@ -1628,24 +1626,33 @@ impl State {
...
-            mutate(|state| {
-                if let Some(user) = state.users.get_mut(&user.id) {
-                    user.treasury_e8s = 0
-                }
-            });
+            .await
+            {
+                mutate(|state| {
+                    if let Some(user) = state.users.get_mut(&user.id) {
+                        user.treasury_e8s = rewards
```
ulan, line 1651: Could `user.treasury_e8s` change in the meantime when we were awaiting the icrc transfer? If so, then this might overwrite that change.

src/backend/env/mod.rs
taggrx: Thanks so much for these comments, Ulan! I rewrote this function completely.

PR #236, opened 2024-01-14
chore: fix recognition of internal links + post rate limiting

taggrx:


chmllr:
LGTM

src/backend/env/post.rs
```diff
@@ -463,31 +463,38 @@ impl Post {
...
-            CONFIG.max_posts_per_hour
-        } else {
-            CONFIG.max_comments_per_hour
-        } as usize;
+        if user
+            .posts(state, 0)
+            .filter(|post| {
+                parent.is_none()
+                    && post.parent.is_none()
+                    && post.timestamp() > timestamp.saturating_sub(DAY)
```
leo-ship-it, line 471: Here I think you are checking for posts that are 1 DAY or older.

src/backend/env/post.rs
```diff
@@ -463,31 +463,38 @@ impl Post {
...
+            ));
+        }
 
         if user
             .posts(state, 0)
             .filter(|post| {
-                !(parent.is_none() ^ post.parent.is_none())
+                parent.is_some()
+                    && post.parent.is_some()
                     && post.timestamp() > timestamp.saturating_sub(HOUR)
```
leo-ship-it, line 487: Here same, aren't you counting posts that are 1 hour or older?

src/backend/env/post.rs
taggrx: No, I think it's 1 DAY or newer: posts, which have a timestamp above `today - one day`, right?

src/backend/env/post.rs
taggrx: No, we're rate limiting. For that I need to know how many posts/comments were produced in the last hour/day, right?

src/backend/env/post.rs
leo-ship-it: You're correct, my bad.

PR #261, opened 2024-01-31
PoW Passwords & principal change confirmation

taggrx:
This PR contains two important changes:

1. It adds a new password based login (by computing the hash 15.000 times which takes about 2-3 seconds on my mac) and an optional migration. The migration is optional meaning that people can skip it, but at every login they will need to approve skipping.
2. The principal change is fail-safe now. Whenever a user asks Taggr to change from principal `A` to principal `B`, the actual change will only happen, if the user logins with the principal `B` after making such a request.

In the future, I plan to change the Taggr backend to an even driven architecture and would even _delay_ the principal change by 24h and notify the user about a pending request.


ulan (commented):
Looks good overall

src/frontend/src/logins.tsx
```diff
@@ -52,17 +64,28 @@ export const authMethods = [
...
-                        ),
-                    );
-                    let serializedIdentity = JSON.stringify(
-                        Ed25519KeyIdentity.generate(hash).toJSON(),
-                    );
+                callback={async (password: string) => {
+                    if (!password) return;
+                    let seed = await hash(password, HASH_ITERATIONS);
+                    let identity = Ed25519KeyIdentity.generate(seed);
+                    const result = await migrateIfNeeded(password);
```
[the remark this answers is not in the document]
taggrx: I thought about it. But then decided against it for these reasons:

1. I'd like to keep the migration code very short and easy to be removed.
2. If people notice that old passwords are fast, no one will migrate 😅

src/frontend/src/logins.tsx
```diff
@@ -153,18 +178,60 @@ export const SeedPhraseForm = ({
...
+        );
+        return oldId;
+    }
+    const newPassword = prompt(
+        "Please specify a new secure password. " +
+            "A secure password should contain at least 8 symbols such as " +
+            "uppercase and lowercase letters, symbols and digits.",
+    );
+    if (!newPassword) return oldId;
+    let seed = await hash(newPassword, HASH_ITERATIONS);
```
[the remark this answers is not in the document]
taggrx: So the prompt asking for the password is actually _not_ hiding it. That means the person typing in the new password will see it in clear text once. I really wasn't sure if a retyping needed then? WDYT?

src/backend/env/mod.rs
```diff
@@ -2606,6 +2616,72 @@ impl State {
...
+            .unwrap_or_default();
+        let all_balances = self.balances.values().sum::<Token>();
+
+        let user_id = self
+            .principals
+            .remove(&principal)
+            .ok_or("no principal found")?;
+        self.principals.insert(new_principal, user_id);
+        let user = self.users.get_mut(&user_id).expect("no user found");
+        user.principal = new_principal;
```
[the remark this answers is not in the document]
taggrx: But the client implementation for this call via the cold wallet does not exist. This is a JS call issued by the frontend (a dfx call will not work). So somebody would need to write hacks to do this for the cold wallet and actually just brick the user, right?

src/backend/env/mod.rs
```diff
@@ -2606,6 +2616,72 @@ impl State {
...
+        user.account = AccountIdentifier::new(&new_principal, &DEFAULT_SUBACCOUNT).to_string();
+        for tx in &mut self.ledger {
+            if tx.to.owner == principal {
```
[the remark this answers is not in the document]
taggrx: Thanks a lot Ulan! Yes, I think it's worth doing! I'll just add a hashmap for principals who migrated and delete it later with the migration code.

src/backend/updates.rs
```diff
@@ -192,12 +192,31 @@ fn update_last_activity() {
...
+
+#[export_name = "canister_update request_principal_change"]
+fn request_principal_change() {
+    let principal: String = parse(&arg_data_raw());
+    mutate(|state| {
+        let principal = Principal::from_text(principal).expect("can't parse principal");
+        if state.principals.contains_key(&principal) {
+            return;
+        }
+        if state.principal_change_requests.len() <= 500 {
```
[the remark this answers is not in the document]
taggrx: Yes, it's a very simple (non-user friendly) way to prevent an unlimited state growth. I'd prefer to keep it dead-simple for now because this lsit of requests will also be reset on every upgrade. Moreover, principal changes are extremely rare. Even with 1M users and an unrealistic assumptions that 1% of users create a request and will not finish, increasing the bounds to 10k will still work fine.

I prefer stupid simple pragmatic solution over user-friendliness in extreme edge cases 😅

src/backend/updates.rs
```diff
@@ -192,12 +192,31 @@ fn update_last_activity() {
...
+}
+
+#[export_name = "canister_update request_principal_change"]
+fn request_principal_change() {
+    let principal: String = parse(&arg_data_raw());
+    mutate(|state| {
+        let principal = Principal::from_text(principal).expect("can't parse principal");
+        if state.principals.contains_key(&principal) {
+            return;
+        }
```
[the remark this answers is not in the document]
taggrx: Yes, thanks! This is simple enough.

src/backend/updates.rs
```diff
@@ -192,12 +192,31 @@ fn update_last_activity() {
...
+        if state.principal_change_requests.len() <= 500 {
+            state.principal_change_requests.insert(principal, caller());
```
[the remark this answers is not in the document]
taggrx: Yes, good idea!

src/backend/env/mod.rs
```diff
@@ -2606,6 +2616,72 @@ impl State {
...
+        if self.voted_on_pending_proposal(principal) {
+            return Err("pending proposal with the current principal as voter exists".into());
+        }
+        let new_principal = Principal::from_text(new_principal_str).map_err(|e| e.to_string())?;
+        if new_principal == Principal::anonymous() {
+            return Err("wrong principal".into());
+        }
+        if self.principals.contains_key(&new_principal) {
+            return Err("principal already assigned to a user".to_string());
+        }
```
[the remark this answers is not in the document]
taggrx: Do you really mean to scan through all txs or just checking the balance?

Second question, I wonder if we want to forbid the following scenario:

1. User uses password A generating principal X. It has txs on the ledger.
2. They change A to B generating a new principal Y.
3. Should no one ever be allowed to use A now? (maybe including the same user?)

src/backend/env/mod.rs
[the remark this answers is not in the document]
ulan: If JS can call it, then dfx can call as well (if you pass a hex encoded JSON as raw argument instead of candid). The attacker can also use js-agent or ic-agent to make the calls programatically. For security, it is better to consider each public endpoint independent from the actual client code.

In this case, if the attacker uses the cold wallet as a the caller, then we would need to prove that nothing can go wrong in the code below. Or we can add a check and rule out that path completely, which simplifies the review and makes the code more robust.

src/backend/env/mod.rs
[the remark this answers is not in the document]
ulan: > Do you really mean to scan through all txs or just checking the balance?

I meant scanning all txs to ensure that the new principal is completely fresh. Alternatively, we could try to prove that nothing goes wrong even if the new principal has some transactions. I think that should work as well.
We could also inline the freshness check in the loop below.

> Second question, I wonder if we want to forbid the following scenario:

After the migration, principal X disappears and becomes fresh, right? It seems okay if someone manages to reuse it. That would be equivalent to guessing the password.

src/frontend/src/logins.tsx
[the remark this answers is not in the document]
ulan: Thinking more about this, it might be worthwhile to bite the bullet and implement a proper form with two password fields instead of using a prompt. The main motivation: the browser's password manager will be able to suggest and save a strong password if it detects the standard password fields. I don't think the password manager will activate in in a prompt.

src/backend/env/mod.rs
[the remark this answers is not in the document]
taggrx: That's a good point!

src/backend/env/mod.rs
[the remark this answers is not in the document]
taggrx: Exactly, but then there is no point in scanning the txs because if even if we find txs in the past, what do we do? IT might be the same user migrates their alt or they guessed a password.

src/frontend/src/logins.tsx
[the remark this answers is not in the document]
taggrx: Will do.

src/backend/env/mod.rs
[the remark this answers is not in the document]
ulan: I meant to check that the new principal `Y` that we are migrating to is completely fresh to make sure that no one is trying to call this endpoint directly (from dfx) using some existing `Y`. I think that should still work, but we need to make sure then that there are no unexpected side-effects.
For example, if the ledger has transactions  `X transferred $N to Y` they will become `Y transferred $N to Y`, which is probably a valid transaction.

src/frontend/src/logins.tsx
```diff
@@ -153,17 +179,96 @@ export const SeedPhraseForm = ({
...
+                value={password2}
+                onChange={(e) => setPassword2(e.target.value)}
+                className="max_width_col top_spaced bottom_spaced"
+                type="password"
+                placeholder="Repeat your password..."
+            />
+            <button
+                className="max_width_col active fat"
+                onClick={() => {
+                    if (password1 != password2) alert("Passwords don't match!");
```
ulan, line 267: Are you intentionally skipping `isSecurePassword(password1)` ? I am on the fence whether we should enforce it. The trade-off here is between guessable and forgettable passwords.

src/frontend/src/logins.tsx
```diff
@@ -153,17 +179,96 @@ export const SeedPhraseForm = ({
...
+}) => {
+    const [password1, setPassword1] = React.useState("");
+    const [password2, setPassword2] = React.useState("");
+    return (
+        <div className="column_container stands_out">
+            <p>
+                Please specify a new secure password. A secure password should
+                contain at least 8 symbols such as uppercase and lowercase
+                letters, symbols and digits.
+            </p>
```
ulan, line 249: With these stricter requirements, there is higher risk of people forgetting the passwords.
I would also mention something along these lines: "Please use the password manager or write down your password on paper and store securely. If you forget your password, your account will be lost. There is no password recovery".

src/frontend/src/common.tsx
```diff
@@ -290,14 +291,18 @@ export const setTitle = (value: string) => {
         ).toUpperCase();
 };
 
+export const HASH_ITERATIONS = MAINNET_MODE ? 400000 : 2;
```
ulan, line 294: Just to double check: did you try this on weaker Android phones? 1 second on iPhone will probably be like 5 second on those.

src/backend/env/mod.rs
```diff
@@ -2034,43 +2038,49 @@ impl State {
...
+        if 0 < balance && balance <= fee as Token {
+            return Err("token balance is lower than the fee".to_string());
+        }
+        let user_id = self
+            .principals
+            .remove(&old_principal)
+            .ok_or("no principal found")?;
+        self.principals.insert(new_principal, user_id);
+        let user = self.users.get_mut(&user_id).expect("no user found");
+        user.principal = new_principal;
```
ulan, line 2067: I would also assert here that `user.principal == old_principal` just to be on the safe side.

ulan (approved):
Lgtm!

src/frontend/src/common.tsx
taggrx: Oh really? No I didn't try it as I have 0 Android devices available to me... Do you have any? I could push this to staging to test there!

ulan: I have an old Pixel 2 to test. If you give me URL, I can try there.

src/backend/env/mod.rs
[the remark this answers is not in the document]
taggrx: I'm confused :D 

> I meant to check that the new principal Y that we are migrating to is completely fresh

My point is that if we do this and find out it is _not_ fresh, then what do we do? If we reject the migration, then we forbid the situation I described in my first response, right? So the main question here is do we want to allow reuse of passwords (accidental or on purpose) or not.

src/frontend/src/logins.tsx
taggrx: Yeah this is also something I thought about longer than I should probably. I don't feel good about _forcing_ users to do something. People who do not understand on their own and are not used to use good passwords, will pick something that they will lose or forget later if we force them to... If they don't and get hacked, is this even DAO's problem anymore? Also note the users who got hacked recently actually accepted that the fault was on their side, because it's kind of obvious, even to them.

src/frontend/src/logins.tsx
taggrx: Thanks, will add!

src/frontend/src/logins.tsx
taggrx: Btw, the difference with the other mask where I do use `isSecurePassword` is because in the login method there are no messages about password strength as we don't know if the user is a new one or not.

src/backend/env/mod.rs
taggrx: Will add!

src/frontend/src/common.tsx
taggrx: Oh this is interesting.

I just deployed this branch to staging. Here's your invite: https://e4i5g-biaaa-aaaao-ai7ja-cai.icp0.io/#/welcome/2a33e03bcb

ulan: I tried 1) sign up 2) sign out 3) sign in.

Both (1) and (3) took about 25 seconds on the old phone.

src/backend/env/mod.rs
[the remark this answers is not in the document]
ulan: Just to double check: the case we are discussing is only about migration, not principal change.

If it is a legitimate migration (user migrating using the client), then we can safely assume that the new principal will be fresh because probability of a hash collision is almost zero.

One case when the new principal is not fresh is if someone is calling this endpoint with an existing principal as a new principal. So I think we can simple reject migration.

Another case where it is possible that the new principal is not fresh and it is not assigned to any user:
1) The user migrates from X to Y
2) The user does some transactions with Y
3) The user changes the principal from Y to Z.
4) The user migrates another account Q to Y (reusing the same password as in step 1).
5) Then we would see Y as not fresh here.
I think it is okay to reject migration here as well.

src/frontend/src/logins.tsx
ulan: Sounds good from my side.

src/frontend/src/common.tsx
taggrx: OH MY GOD. But imo setting it to 2-3 seconds on an old device will make it instant on a Sillicon Mac... 🤔

ulan: Yeah, it will make it 10x faster, but it is still 40000x slower compared to the previous state? Still sounds okay, no?
Did you measure how many milliseconds it takes on Silicon?

Note that my Android phone is quite old. We could also try to get other sample from community?

taggrx: Yes! So right now we have 300k iterations on Staging. It sounds like 10k could work. I'll set it to 10k and then ask people to confirm.

ulan: Sounds good! Another thing to mention is that we are not using a [salt](https://en.wikipedia.org/wiki/Salt_(cryptography)). I guess it is because it would complicate the flow because we would need to store the salt in the backend and users would need to query it before login. So we are trading security for simplicity of code here. Just wanted to make sure that this is a conscious decision.

src/backend/env/mod.rs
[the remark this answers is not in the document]
taggrx: > I think it is okay to reject migration here as well.

Ok, that was basically my question all the time. I have no hard feels, so if you feel like we should forbid it, let's do it. 

The only thing I don't like is scanning the ledger because at some point it will be very expensive. However, I think we can remove the migration in half a year or so and until then I do not expect the ledger to explode, so I think we're fine.

taggrx:
@ulan  I also added a counter of successful migrations on the state level to be able to monitor if people keep migrating. Then we know where it can be removed.

src/backend/env/mod.rs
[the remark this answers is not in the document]
ulan: Gotcha. I was on the edge here thinking that it is not that bad if we allow migrating to an existing principal as long as we prove that nothing goes wrong. I tried to prove it now and found a bad case.

Let's say someone is a bad actor and has done some bad transactions. Currently, that bad principal can "frame/taint" a good principal of someone else by calling migrate to that good principal. The endpoint doesn't check that the caller controls the new principal.

src/frontend/src/common.tsx
taggrx: And because as discussed before, a salt stored in the state which can be fully exported is not adding security. Only an extra step to the hacker :D

ulan: Even if the salt is public, it is still useful against dictionary attacks. Without salts, what the attacker can do is download a database of leaked/common passwords and then for each password run the hashing algorithm on cluster of machines saving a dictionary of `principal => password`. Then go over all known principals to see matches. So it is like run the hashing once for each password and re-use it for many principals.

With a salt this attack no longer works because for each principal/salt the attacker needs to brute force the entire password set.

taggrx: Oh, you're talking about one salt per user? But this requires an additional identification anchor to know which user Taggr talks to. Otherwise it wouldn't know for which user the salt should be pulled from the backend before computing the seed and hence principal which is currently identifying the user. Or am I missing something?

I should have never renamed it to "password" login and actually insist on entering BIP39 phrases... 😅

ulan: Right, that what meant by "would complicate the flow because we would need to store the salt in the backend and users would need to query it before login". One possibility would be to query the backend based on the user name and get a list of possible salts to try on the client side and then query the backend again to check if the principals exist... I guess, let's not go there.

taggrx: Yes. Imo passwords will ultimately always be less secure because they will always be susceptible to key loggers, 5$ wrench attacks and what not. I'd rather investigate how to add passkeys in the future.

PR #271, opened 2024-02-11
Optimize image loading

ulan:
Currently images are loaded as opaque blobs via a query call to the bucket canister.
The images are later converted to URLs with `createObjectURL`.

This PR replaces blobs with proper images that are served by the bucket canister via
the HTTP protocol (`http_request` query).

This has multiple performance benefits:
- the browser knows that it is loading images and can optimise it.
- the browser can cache the images.
- the boundary node can cache the images.
- the browser doesn't waste time in encoding blobs as URLs.
- the browser makes an anonymous request which allows both the boundary node
   and the replica to share the caches between multiple users.

Local testing results on a page with many images and with 3G network throttling:
- Old code, first load: 42s,
- Old code, second load: 43s,
- New code, first load: 26s,
- New code, second  load: 14s (all this time is spent loading
  non-image resources, images are served from the disk cache).

Summary of the changes:
- The bucket backend now has a new `http_request` endpoint that  serves 
  `/image?offset=X&len=Y` queries with a cache-control header.
- The frontend uses `urls` instead of `blobs`.
- The new post form converts blobs to URLs.


ulan:
> Old code, first load: 42s,
> New code, first load: 26s,

I don't fully understand why the first load improved so much. It doesn't seem to be a flake since I got the same results a few times.

ulan:
> > Old code, first load: 42s,
> > New code, first load: 26s,
> 
> I don't fully understand why the first load improved so much. It doesn't seem to be a flake since I got the same results a few times.

Based on the videos:
- https://www.youtube.com/watch?v=YKNEaMd4TRg
- https://www.youtube.com/watch?v=Cg1IHTBBOkk

The new code downloads 2.8MB whereas the old code download 6.6MB. I wonder if the additional overhead comes from the CBOR encoding that's done for queries.

src/frontend/src/common.tsx
```diff
@@ -628,36 +629,34 @@ export const BurgerButton = ({
...
-            const api = window.backendCache.stats.buckets.every(
+            const url = window.backendCache.stats.buckets.every(
                 ([id]) => id != bucket_id,
             )
-                ? window.mainnet_api
-                : window.api;
-            return api
-                .query_raw(bucket_id, "read", Buffer.from(args))
-                .then((blob) => [blobId, blob || new ArrayBuffer(0)]);
+                ? ""
```
ulan, line 644: Is this okay? I don't understand this special case.

src/frontend/src/common.tsx
```diff
@@ -628,36 +629,34 @@ export const BurgerButton = ({
...
+                : bucket_image_url(bucket_id, offset, len);
+            return new Promise((resolve, reject) => {
```
ulan, line 646: I also tried skipping this pre-loading and simply returning the url here. It causes the post heights to change while images are loading, which is probably not a good user experience.

src/bucket/src/lib.rs
```diff
@@ -36,18 +61,68 @@ fn balance() {
     reply_raw(&canister_balance().to_be_bytes())
 }
 
+#[ic_cdk_macros::query]
+fn http_request(req: HttpRequest) -> HttpResponse {
+    let url = url::parse(&req.url);
+    match url.path {
+        "/image" => http_image(url.args),
```
ulan, line 68: I can also change this to `/blob` if there is a preference. I chose `/image` because we are returning the `image/png` content type.

leo-ship-it:
Great idea Ulan, dividing by 2 load times is really impressive!

src/bucket/src/lib.rs
```diff
@@ -1,10 +1,35 @@
...
     stable::*,
 };
+use serde::Serialize;
+use serde_bytes::ByteBuf;
+
+mod url;
+
+// An upper bound on the blob size. Queries above this size will be rejected
+// without trying to read the stable memory.
+const MAX_BLOB_SIZE: u64 = 8 * 1024 * 1024;
```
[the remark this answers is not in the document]
ulan: Yes, it is 2MB. IIRC, in there are some exceptional cases that may allow 3MB (I think this the system subnet).

In any case, this constant doesn't need to be precise. We just need some number that's definitely higher than IC's limit and low-enough to avoid wasted work.

src/bucket/src/lib.rs
```diff
@@ -1,10 +1,35 @@
...
+    headers: Headers,
+}
+
+#[derive(CandidType, Serialize)]
+struct HttpResponse {
+    status_code: u16,
+    headers: Headers,
+    body: ByteBuf,
+    upgrade: Option<bool>,
+}
```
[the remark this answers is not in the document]
ulan: Thanks for the suggestion! Let's check if @taggrx wants this dependency. (I copied these definitions from the backend)

ulan:
There is a bug in editing, that's why the test is failing. I am fixing it right now.

src/bucket/src/lib.rs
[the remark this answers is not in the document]
leo-ship-it: Then this constant seems reasonable.

src/frontend/src/form.tsx
```diff
@@ -90,19 +93,23 @@ export const Form = ({
             return false;
         }
         setSubmitting(true);
-        const blobArrays = Object.keys(tmpBlobs).reduce(
-            (acc, id) => {
-                if (value.includes(`(/blob/${id})`)) {
-                    // @ts-ignore
-                    acc.push([id, [...tmpBlobs[id]]]);
```
ulan, line 97: @taggrx: what was the intention behind `...tmpBlobs[id]`?

I replaced it with `tmpBlobs[id]`, which I think should also work.

src/frontend/src/common.tsx
taggrx: This displays mainnet images on a local instance. Sometimes it helps with debugging of UI issues. So what happens here is if a picture is referenced from a bucket we don't have in our config, we assume it's the mainnet bucket. But I think the issue is automatically fixed if we use the URLs right?

src/frontend/src/content.tsx
```diff
@@ -125,9 +123,7 @@ export const Content = ({
             <>
                 {markdownizer(
                     value,
-                    urls,
-                    setUrls,
-                    blobs,
+                    urls ? urls : {},
```
taggrx, line 126:
```suggestion
                    urls || {},
```

src/frontend/src/content.tsx
```diff
@@ -138,16 +134,14 @@ export const Content = ({
                     ) : (
                         markdownizer(
                             extValue,
-                            urls,
-                            setUrls,
-                            blobs,
+                            urls ? urls : {},
```
taggrx, line 137:
```suggestion
                            urls || {},
```

src/frontend/src/env.ts
```diff
@@ -11,3 +11,14 @@ export const II_URL = MAINNET_MODE
 export const II_DERIVATION_URL = MAINNET_MODE
     ? `https://${CANISTER_ID}.ic0.app`
     : window.location.origin;
+
+export function bucket_image_url(
+    bucket_id: string,
+    offset: number,
+    len: number,
+) {
+    let host = MAINNET_MODE
```
taggrx, line 20: I think to make it work as it is right now, we can simply replace `MAINNET_MODE` with something like `!window.backendCache.stats.buckets.includes(bucket_id)`?

So that if the bucket is not in the local config, we just try to load from mainnet?

src/frontend/src/form.tsx
taggrx: I can't remember either. You change LGTM!

taggrx (commented):
Thanks so much Ulan! I can't wait to roll it out! 😅

This is the first review round. Once you apply/reject/challenge the suggestions, I'll do another one, by debugging how things work on my local replica!

src/bucket/Cargo.toml
```diff
@@ -8,7 +8,10 @@ crate-type = ["cdylib"]
 
 [dependencies]
 ic-cdk = "0.11.3"
-candid = "0.9.11"
+ic-cdk-macros = "0.8.1"
+candid = { version = "0.9.11", features = ["parser"] }
```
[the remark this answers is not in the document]
ulan: Good point! I don't remember why I changed it. It works without the parser. Removed.

src/frontend/src/env.ts
ulan: Added `MAINNET_MODE || fallback_to_mainnet`

ulan (commented):
Thanks for the review! I applied the comments adjusting for compile errors where needed.

src/frontend/src/common.tsx
ulan: I removed the conditional here because the url helper function now handles this case.

ulan:
I am signing off for today. Please feel free to push changes as you see fit.. Or merge and follow up with your fixes.

PR #274, opened 2024-02-13
Image caching

taggrx:
The image uploading does not work for me yet (neither locally, nor on staging - image queries return 404). I'll investigate later.


taggrx:
@ulan so I could greatly simplify the code even further as no lazy loading needs to be re-implemented on JS side and can be outsourced to the browser.

However, I am stuck on the issue I had from the beginning with your change: the images simply do not load. A concrete example: on my local replica the image referenced as URL `http://gx2xg-kmaaa-aaaaa-qaasq-cai.raw.localhost:8080/image?offset=28520&len=28512` does not load anything and displays `Cannot GET /image` (whatever that means) if opened in a new tab. But if I call the bucket via DFX, I see the expected response:

    $ dfx canister call --query gx2xg-kmaaa-aaaaa-qaasq-cai http_request '(record {url = "/image?offset=28520&len=28512"; headers = vec {} })'

The response is the candid response from the canister. So I first assumed that it's the local boundary node emulation (or whatever this thing is) that can't process the HTTP requests. So I deployed the branch to staging, and there I see the same issue. For example, [this URL](https://hztov-cqaaa-aaaao-a2iiq-cai.raw.ic0.app/image?offset=9127034&len=28512) from [this post](https://e4i5g-biaaa-aaaao-ai7ja-cai.icp0.io/#/post/213) returns 404. But also the DFX call does not work and I don't even understand the error message:

```
$ dfx canister --network ic call --query hztov-cqaaa-aaaao-a2iiq-cai http_request '(record {url = "/image?offset=9127034&len=28512"; headers = vec{}})'
Error: Failed query call.
Caused by: Failed query call.
  The replica returned an HTTP Error: Http Error: status 400 Bad Request, content type "text/plain; charset=utf-8", content: unable_to_parse_http_arg: unable to decode arg as HttpRequest: input: 4449444c026c02efd6e40271c6a4a19806016d6f01001f2f696d6167653f6f66667365743d39313237303334266c656e3d3238353132_00
table: type table0 = record { 5_843_823 : text; 1_661_489_734 : table1 }
type table1 = vec empty
wire_type: text, expect_type: text, field_name: Named("method")
```

However it works if I remove the `--query` parameter... 😳 🤔 

I'm completely lost right now and would appreciate any hints or suggestions! 🙏

taggrx:
Ok, I could fix the staging by using `icp0.io` instead of the `ic0.app` as a host name.

taggrx:
Locally everything started to work after restarting the local stack...

PR #275, opened 2024-02-13
bump deps

leo-ship-it:
bump candid and related deps to latest to check candid interface.


leo-ship-it:
My plan is to get rid of the usage of the "old-school" candid.
- Move from `parse` to params.
- Move from `export_name` to ic_cdk_macros.

taggrx:
Thanks so much Leo for getting rid of the parser feature! 🚀

PR #277, opened 2024-02-14
amended proposal prompt to help finding the hash

hannahaddad:
a tiny commit but hopefully it will help new users. that hashtag is marked on one of my posts that shows you how to get the hash. simplest change for the least effort considering that text is in a window prompt.


taggrx:
Thanks a lot Hanna! However, don't you think that adding references to app data from app _code_ is not only error prone (data is volatile) but can also be dangerous. Imagine, somebody adds a new post with this tag with some malicious intent which leads to rugging people?

hannahaddad:
> Thanks a lot Hanna! However, don't you think that adding references to app data from app _code_ is not only error prone (data is volatile) but can also be dangerous. Imagine, somebody adds a new post with this tag with some malicious intent which leads to rugging people?

You make a valid point. let me think of an alternative.

PR #281, opened 2024-02-14
refactor: don't parse args on updates

leo-ship-it:
I started with this one to see what you think @taggrx.

Is there any valid reason to not show some endpoints in the did file? I set to `hidden` the endpoints that were not present in the did file.

Could I submit a proposal to upgrade? If so, do you have a guide on how to submit an upgrade, or should I just figure out myself by looking at the code. I guess you just call `propose_release`.

Based on your feedback I'll also update the queries a get rid of the parse method.


taggrx:
Leo I think you misunderstood me on OpenChat. What you're trying to do **does not work**. All these endpoints are JSON endpoints. They have nothing to do with Candid and do not work with the macros provided by the CDK. Look at how they're called: https://github.com/TaggrNetwork/taggr/blob/main/src/frontend/src/api.ts#L142

leo-ship-it:
Ok, got it. Is there a specific reason to use JSON instead of candid?

taggrx:
> Ok, got it. Is there a specific reason to use JSON instead of candid?

Yes: there is no sense in using Candid. It comes with **zero** benefits for backend <> frontend communication.

JSON is native to web development and is a breeze to use. For example, open the JS console on Taggr and and enter: `api.query("user", ["leo"]).then(console.log)`

See how cool it is? Now good luck with doing anything like this with Candid (after you spent an hour debugging and realizing that the auto-generated JS code is in not sync with your backend code and you need to restart the entire stack 10 times to get it in sync).

I switched to JSON 2 years ago and never looked back.

leo-ship-it:
I see, thanks for the explanation!

PR #282, opened 2024-02-15
Release

taggrx:
- image caching by Ulan
- minor fixes


src/backend/updates.rs
```diff
@@ -70,7 +70,9 @@ fn post_upgrade() {
     });
 }
 
-async fn post_upgrade_fixtures() {}
+async fn post_upgrade_fixtures() {
+    env::storage::upgrade_buckets().await;
```
taggrx, line 74: We have to add this to upgrade the bucket.

